PR #6, opened 2017-07-31
Explaining why Crossbeam is correct w.r.t. C11 memory model

jeehoonkang:
Here is the promised RFC on the concurrency memory model and correctness of Crossbeam.

[Rendered](https://github.com/jeehoonkang/crossbeam-rfcs/blob/relaxed-memory/text/2017-07-23-relaxed-memory.md)

### DIFF
--- a/text/2017-07-23-relaxed-memory.md
+++ b/text/2017-07-23-relaxed-memory.md
@@ -45,9 +45,9 @@ w.r.t. the C11 memory model.
 ## Proposed implementation
 
 As discussed in Aaron Turon's original [blog post][aturon]
-and [the pull request of a previous RFC][rfc2-pr], Crossbeam guarantees that *every access to a
+and [the pull request of a previous RFC][rfc2-pr], Crossbeam guarantees that *every access to an
 object should happen before the object is deallocated*.  In order to provide this guarantee,
-Crossbeam assumes that when a pinned thread marks a object as unlinked, *the thread already removed
+Crossbeam assumes that when a pinned thread marks an object as unlinked, *the thread already removed
 all the references to the object from the memory on the thread's point of view.*; and Crossbeam
 defers the deallocation of an unlinked object until all the threads concurrently pinned at the time
 the object is unlinked, are unpinned.
@@ -56,7 +56,7 @@ In order to track the concurrently pinned threads, Crossbeam utilizes the epoch 
 
 - There is a local epoch `th.epoch` for each *pinned* thread `th`, and a global epoch `EPOCH`.
 
-- When a object is deallocated, it is marked with the **current** global epoch, as in the following
+- When an object is deallocated, it is marked with the **current** global epoch, as in the following
   pseudocode (for now, please ignore the fences and the orderings such as `Relaxed`, `Acquire`,
   `Release`, and `SeqCst`):
 
@@ -95,13 +95,13 @@ In order to track the concurrently pinned threads, Crossbeam utilizes the epoch 
   ```rust
   fn try_advance() {
     'L40: let G = EPOCH.load(Relaxed);
-    'L41: fence(SeqCst);
+    'L41: atomic::fence(SeqCst);
 
     'L42: for th in threads {
     'L43:   let e = th.epoch.load(Relaxed);
     'L44:   if e != SENTINEL && e != G { return; }
     'L45: }
-    'L46: fence(Acquire);
+    'L46: atomic::fence(Acquire);
 
     'L47: EPOCH.store(G + 1, Release);
 
@@ -111,7 +111,7 @@ In order to track the concurrently pinned threads, Crossbeam utilizes the epoch 
 
 Note that at `pin(): 'L23` and `try_advance(): 'L48`, an unlinked object marked with an epoch two
 generations before than the current global one, can be deallocated.  In other words, you can
-deallocate a object marked with `X` when the current global epoch `>= X+2`.  This is because:
+deallocate an object marked with `X` when the current global epoch `>= X+2`.  This is because:
 
 - For each pinned thread `th`, as an invariant, the local epoch `th.epoch` equals to either `EPOCH`
   or `EPOCH-1`.  Thus when `EPOCH >= X+2`, a local epoch should be greater than or equal to `X+1`.
@@ -127,7 +127,7 @@ deallocate a object marked with `X` when the current global epoch `>= X+2`.  Thi
 
 Now we explain why the proposed implementation is correct w.r.t. the C11 memory model.
 
-Suppose that the current epoch is `X+2`, and a object `unlink()`ed and marked with `X` is about to
+Suppose that the current epoch is `X+2`, and an object `unlink()`ed and marked with `X` is about to
 be deallocated.  For correctness, we need to ensure that a concurrent `pin()`ned thread cannot
 access the object.
 
@@ -156,13 +156,13 @@ memory on its point of view.  By cumulativity, `pin()`ned thread cannot access t
 - By transitivity, `pin()`'s SC fence is performed before `try_advance()`'s SC fence.  Thus the
   store to the local epoch at `pin(): 'L21` is SC-executed before the load from it at
   `try_advance(): 'L43`.  Thus `'L43` should read from what is written at `'L21` or a more recent
-  value that.
+  value than that.
 
 - In order to reach `try_advance(): 'L47` and increment the global epoch, `'L43` should not read
   from what is written at `'L21`.  It should read from what is written at `unpin(): 'L30` or a more
   recent value than that.  Thus, by release-acquire synchronization, `'L30` happens before `'L46`.
   
-- Since a thread should access a object before `unpin()`, every access to the unlinked object from
+- Since a thread should access an object before `unpin()`, every access to the unlinked object from
   the `pin()`ned thread happens before the object's deallocation.
 
 
--- a/text/2017-07-23-relaxed-memory.md
+++ b/text/2017-07-23-relaxed-memory.md
@@ -9,10 +9,10 @@ w.r.t. C11/C++11/LLVM/Rust relaxed-memory concurrency model (the "C11 memory mod
 # Motivation
 
 Crossbeam is based on the epoch-based memory reclamation scheme (EBR).  In EBR, an **epoch** (think:
-timestamp) is maintained, and a location is mark with the current epoch when the location is
-unlinked.  The unlinked location is safe to deallocate later when the epoch is advanced enough.  As
-you may guess, the correctness of EBR crucially depends on the invariant on the epoch: it is
-advanced only when old locations unlinked in the past are no longer accessible.
+timestamp) is maintained, and an object is mark with the current epoch when the object is unlinked.
+The unlinked object is safe to deallocate later when the epoch is advanced enough.  As you may
+guess, the correctness of EBR crucially depends on the invariant on the epoch: unlinked objects are
+inaccessible after two advancements.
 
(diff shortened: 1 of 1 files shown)


text/2017-07-23-relaxed-memory.md
```diff
@@ -0,0 +1,218 @@
...
+This is a proposal for clarifying the synchronization performed in the epoch-based memory
+reclamation scheme, and explaining why Crossbeam's implementation is correct
+w.r.t. C11/C++11/LLVM/Rust relaxed-memory concurrency model (the "C11 memory model" from now on).
+
+
+
+# Motivation
+
+Crossbeam is based on the epoch-based memory reclamation scheme (EBR).  In EBR, an **epoch** (think:
+timestamp) is maintained, and a location is mark with the current epoch when the location is
```
ghost: Instead of "location" I'd prefer "object", "node", "garbage", or something similar.
Or maybe "memory location", or "memory allocation"?

text/2017-07-23-relaxed-memory.md
```diff
@@ -0,0 +1,218 @@
...
+unlinked.  The unlinked location is safe to deallocate later when the epoch is advanced enough.  As
+you may guess, the correctness of EBR crucially depends on the invariant on the epoch: it is
+advanced only when old locations unlinked in the past are no longer accessible.
```
ghost: Well, the invariant is more like "unlinked objects are inaccessible after *two* advancements", isn't it?

text/2017-07-23-relaxed-memory.md
```diff
@@ -0,0 +1,218 @@
...
+## Non-motivation
+
+Crossbeam is the marriage of EBR and Rust's linear type system with lifetime.  However, this RFC
+does not aim for clarifying how Rust's type system helps to ensure the correctness of Crossbeam.
+
+
+
+# Detailed design
+
+In this section, we present the proposed implementation of Crossbeam, and explain why it is correct
```
ghost: "explain why it is correct" - technically, isn't this a proof? Can we say that we "prove it is correct", or "show a sketch of proof of correctness"?

text/2017-07-23-relaxed-memory.md
```diff
@@ -0,0 +1,218 @@
...
+
+- There is a local epoch `th.epoch` for each *pinned* thread `th`, and a global epoch `EPOCH`.
+
+- When a location is deallocated, it is marked with the **current** global epoch, as in the
+  following pseudocode (for now, please ignore the fences and the orderings such as `Relaxed`,
+  `Acquire`, `Release`, and `SeqCst`):
+
+  ```rust
+  fn unlink(l) {
+    'L10: atomic::fence(SeqCst);
```
ghost: Do we really need this fence - can we just have a `SeqCst` load instead? I think that'd be still correct.

ghost (changes requested):
Looks very good - I'm excited to see progress on this front!

I have a few comments on the clarity of the text, and wonder whether we could replace a `SeqCst` fence with a `SeqCst` operation.

Also, the proof could benefit from a drawing of some kind of timeline of atomic fences/operations (ASCII art would be just fine), since there's a lot of stuff going on.

text/2017-07-23-relaxed-memory.md
jeehoonkang: I would rather not call it a proof, as there are many gaps..  For concurrency, without actually formallying  proving it, I am hesitant to say it is a proof..

text/2017-07-23-relaxed-memory.md
jeehoonkang: I briefly described why it is insufficient in the "current implementation" subsection in the "alternatives" section.  That is, `SeqCst` load lacks "cumulativity", which is essential in the above explanation.

jeehoonkang:
I addressed @stjepang 's comments.  Thank you!

Argh, I omitted this comment:

> Also, the proof could benefit from a drawing of some kind of timeline of atomic fences/operations (ASCII art would be just fine), since there's a lot of stuff going on.

I will do it this week.

text/2017-07-23-relaxed-memory.md
ghost: I think I understand now. Thanks.

Just a side note (an alternative): We could use a `Relaxed` load here without the fence, but then destroy garbage when `G >= X + 3` rather than `G >= X + 2`.

text/2017-07-23-relaxed-memory.md
ghost: Fair enough. :)

text/2017-07-23-relaxed-memory.md
jeehoonkang: Could you explain a bit further why it works?  I feel that it will work, but I want to understand it better.

I think it is an interesting idea.  SC fence is expensive, so it may be worth removing at the cost of delayed garbage collection.  I want to think of why it should be correct, and see a benchmark result.

jeehoonkang:
I added an ascii art on an example timeline.  Thank you!

text/2017-07-23-relaxed-memory.md
ghost: Sure. Consider your proposed implementation:

```rust
fn unlink() {
    atomic::fence(SeqCst);
    let X = epoch.load(Relaxed);
    // ...
}
```

And the alternative:

```rust
fn unlink() {
    let Y = epoch.load(Relaxed);
    // ...
}
```

Also, let's say that `G_pin` is the global epoch just after the current thread was pinned:

```rust
fn pin(th) {
    let e = EPOCH.load(Relaxed);
    th.epoch.store(e, Relaxed);
    atomic::fence(SeqCst);
    let G_pin = EPOCH.load(Relaxed);
}
```

Note that:

1. The current thread is pinned while `unlink` is called.
2. `G_pin <= X <= G_pin + 1`, because the global epoch can advance at most once while our thread is pinned.
3. `G_pin <= Y`, because `G_pin` was loaded before `Y`.
4. From 2. and 3. we can derive: `Y >= X - 1`.

The invariant is that garbage tagged with `X` can be destroyed if `G >= X + 2`.
Or, equivalently, garbage can be destroyed if `G >= X - 1 + 3`.

Since `Y + 3 >= X - 1 + 3`, we can be slightly **more conservative** and say:
Garbage can be destroyed if `G >= Y + 3`.
That is a slightly stronger inequality than `G >= X + 2`.

And of course, computing `G_pin` is not necessary - that is just to show why `G >= Y + 3` works. :)

I think it's worth remembering that tagging garbage without a `SeqCst` is perfectly possible with a very minor tweak. We'll benchmark that later.

arthurprs:
Nice work! I'm just coming back from vacations and I plan to review this soon.

ghost:
I gave the text another, more detailed read. As far as I'm concerned, it's almost good to go.

Let's just add another subsection to *Alternatives* and mention that it's possible to eliminate the `SeqCst` fence in `unlink()` by changing the invariant from `G >= X+2` to `G >= X+3`.

schets:
> Let's just add another subsection to Alternatives and mention that it's possible to eliminate the SeqCst fence in unlink() by changing the invariant from G >= X+2 to G >= X+3.

I haven't had time to read this in detail yet, but if this statement is true, we should almost certainly do this. On x86, this is the only type of fence that results in a hardware fence and it's a performance disaster, and on ArmV8 and Power this fence is more expensive than acquire/release fences.

text/2017-07-23-relaxed-memory.md
jeehoonkang: For the case that the a concurrent thread's `pin()`'s SC fence is executed before the unlinking thread's `pin()`'s SC fence, your algorithm seems right.  However, I am not sure whether it works for the case that the SC fences are executed in the reverse order (i.e. the unlinking thread's `pin()`'s SC fence is executed first).

arthurprs:
>  it's possible to eliminate the SeqCst fence in unlink() by changing the invariant from G >= X+2 to G >= X+3.

Very interesting. Did you toyed with that yet? This potentially makes pinning much cheaper on non x86 (as x86 can use a lock CAS).

ghost:
> Very interesting. Did you toyed with that yet? This potentially makes pinning much cheaper on non x86 (as x86 can use a lock CAS).

Did you mean *unlinking* instead of *pinning*?

I didn't toy with this much, but: in Coco garbage is tagged with the current epoch only when flushing thread-local bags into the global queue. The whole bag is tagged with the current epoch. Since tagging is so infrequent, its effect on performance is negligible.

In Crossbeam we might employ a slightly different strategy and tag garbage more often, but it's still probably unnecessary to tag absolutely every garbage object with the current epoch. For that reason eliminating the `SeqCst` fence possibly won't get us any noticeable performance wins...

arthurprs:
Indeed, I got it all wrong. I need extra :coffee: before looking into this thread .

jeehoonkang:
I mentioned @stjepang's proposal (to remove the SC fence from `unlink()`) in the "Unresolved Questions" section, because I am still not fully convinced that the proposal is correct.  (I agree with @schets on the performance of the proposal, even considering the fact that `unlink()`'s SC fences are amortized.)  I suggest we continue the discussion of https://github.com/crossbeam-rs/rfcs/pull/6#discussion_r130942541 in another RFC.

text/2017-07-23-relaxed-memory.md
```diff
@@ -0,0 +1,240 @@
...
+    'L40: let G = EPOCH.load(Relaxed);
+    'L41: atomic::fence(SeqCst);
+
+    'L42: for th in threads {
+    'L43:   let e = th.epoch.load(Relaxed);
+    'L44:   if e != SENTINEL && e != G { return; }
+    'L45: }
+    'L46: atomic::fence(Acquire);
+
+    'L47: EPOCH.store(G + 1, Release);
```
arthurprs, line 107: I think L47 is correct due to the logic around the epochs and insurance that the current thread is also registered. Otherwise it needs to be a CAS right?

jeehoonkang: Exactly.  Multiple thread may concurrently write to rhe global epoch the same value, but that write-write race doesn't matter thanks to SC fences.

ticki (approved):

ghost (approved):

arthurprs (approved):
:tada:

jeehoonkang:
Thank you for all the inputs so far.  @schets @Vtec234 may I ask if you have further comments?

Vtec234 (approved):
Sorry for the delay in reviewing, I've been a little busy recently. Looks correct from what I've read, but I'll give it a more detailed reading later. Besides showing why freeing a pointer from two epochs ago is correct, it would also be great to show why a pointer that is currently in use cannot be freed.

jeehoonkang:
> it would also be great to show why a pointer that is currently in use cannot be freed.

@Vtec234 Thank you for the suggestion.  I slightly revised the text so that it clearly explains that all the accesses to `obj` should happen before its deallocation.

text/2017-07-23-relaxed-memory.md
```diff
@@ -0,0 +1,246 @@
...
+
+    'L48: // now objects marked with `<= e - 1` can be deallocated.
```
Vtec234, line 109: Is this correct? `pin` mentions `e - 2` rather than `e - 1`

jeehoonkang: Yes, since now the global epoch is `G+1 = e+1`, and garbages marked with `(G+1)-2 = e-1` can be safely deallocated.

Vtec234: Ah yeah of course, I can't do basic math

jeehoonkang:
I am merging it based on 4 approvals :)  Thank you for all the comments!

PR #9, opened 2017-08-30
Simplify deferred destruction

ghost:
Proposal:

* Remove `Scope::defer_free`.
* Optimize `Scope::defer`.

[Rendered](https://github.com/stjepang/rfcs-crossbeam/blob/simplify-defer/text/2017-08-30-simplify-defer.md)

### DIFF
--- a/text/2017-08-30-simplify-defer.md
+++ b/text/2017-08-30-simplify-defer.md
@@ -121,12 +121,13 @@ Put differently, `Deferred` is to `FnOnce() + Send + 'static` what `SmallVec` is
 
 <details>
 <summary>Code (click to expand)</summary>
+
 ```rust
 use std::mem;
 use std::ptr;
 
 /// Provides methods to dispatch a call to a `FnOnce()` from a trait object.
-trait Callback {
+pub trait Callback {
     /// Calls the function from a trait object on the stack.
     ///
     /// This will copy `self`, call the function, and finally drop the copy.
@@ -170,30 +171,24 @@ struct TraitObject {
 }
 
 /// Some space to keep a `FnOnce()` object on the stack.
-type Data = [u64; 4];
-
-/// A small `FnOnce()` stored inline on the stack.
-struct InlineObject {
-    data: Data,
-    vtable: *mut (),
-}
+type Data = [usize; 3];
 
 /// A `FnOnce()` that is stored inline if small, or otherwise boxed on the heap.
 ///
 /// This is a handy way of keeping an unsized `FnOnce()` within a sized structure.
-enum Deferred {
-    OnStack(InlineObject),
-    OnHeap(Option<Box<Callback>>),
+pub struct Deferred {
+    vtable: *mut (),
+    data: Data,
 }
 
 impl Deferred {
     /// Constructs a new `Deferred` from a `FnOnce()`.
-    fn new<F: FnOnce() + Send + 'static>(f: F) -> Self {
+    pub fn new<F: FnOnce() + Send + 'static>(f: F) -> Self {
         let size = mem::size_of::<F>();
         let align = mem::align_of::<F>();
 
-        if size <= mem::size_of::<Data>() && align <= mem::align_of::<Data>() {
-            unsafe {
+        unsafe {
+            if size <= mem::size_of::<Data>() && align <= mem::align_of::<Data>() {
                 let vtable = {
                     let callback: &Callback = &f;
                     let obj: TraitObject = mem::transmute(callback);
@@ -201,38 +196,37 @@ impl Deferred {
                 };
 
                 let mut data = Data::default();
-                ptr::copy_nonoverlapping(
-                    &f as *const F as *const u8,
-                    &mut data as *mut Data as *mut u8,
-                    size,
-                );
-                mem::forget(f);
+                ptr::write(&mut data as *mut Data as *mut F, f);
 
-                Deferred::OnStack(InlineObject { data, vtable })
+                Deferred { vtable, data }
+            } else {
+                let mut data = Data::default();
+                let b: Box<Callback> = Box::new(f);
+                ptr::write(&mut data as *mut Data as *mut Box<Callback>, b);
+
+                Deferred {
+                    vtable: 0x1 as *mut (),
+                    data,
+                }
             }
-        } else {
-            Deferred::OnHeap(Some(Box::new(f)))
         }
     }
 
     /// Calls the function or panics if it was already called.
     #[inline]
-    fn call(&mut self) {
-        match *self {
-            Deferred::OnStack(ref mut obj) => {
-                let vtable = mem::replace(&mut obj.vtable, ptr::null_mut());
-                assert!(!vtable.is_null(), "cannot call `FnOnce` more than once");
+    pub fn call(&mut self) {
+        let vtable = mem::replace(&mut self.vtable, ptr::null_mut());
+        assert!(!vtable.is_null(), "cannot call `FnOnce` more than once");
 
-                unsafe {
-                    let data = &mut obj.data as *mut _ as *mut ();
-                    let obj = TraitObject { data, vtable };
-                    let callback: &Callback = mem::transmute(obj);
-                    callback.copy_and_call();
-                }
-            }
-            Deferred::OnHeap(ref mut opt) => {
-                let boxed = opt.take().expect("cannot call `FnOnce` more than once");
-                boxed.call_box();
+        unsafe {
+            if vtable as usize != 0x1 {
+                let data = &mut self.data as *mut _ as *mut ();
+                let obj = TraitObject { data, vtable };
+                let callback: &Callback = mem::transmute(obj);
+                callback.copy_and_call();
+            } else {
+                let b: Box<Callback> = ptr::read(&self.data as *const Data as *const Box<Callback>);
+                b.call_box();
             }
         }
     }
@@ -244,14 +238,14 @@ mod tests {
 
     #[test]
     fn smoke_on_stack() {
-        let a = [0u64; 1];
+        let a = [0usize; 1];
         let mut d = Deferred::new(move || drop(a));
         d.call();
     }
 
     #[test]
     fn smoke_on_heap() {
-        let a = [0u64; 10];
+        let a = [0usize; 10];
         let mut d = Deferred::new(move || drop(a));
         d.call();
     }
@@ -259,7 +253,7 @@ mod tests {
     #[test]
     #[should_panic(expected = "cannot call `FnOnce` more than once")]
     fn twice_on_stack() {
-        let a = [0u64; 1];
+        let a = [0usize; 1];
         let mut d = Deferred::new(move || drop(a));
         d.call();
         d.call();
@@ -268,7 +262,7 @@ mod tests {
     #[test]
     #[should_panic(expected = "cannot call `FnOnce` more than once")]
     fn twice_on_heap() {
-        let a = [0u64; 10];
+        let a = [0usize; 10];
         let mut d = Deferred::new(move || drop(a));
         d.call();
         d.call();
@@ -287,8 +281,16 @@ mod tests {
         let mut d = Deferred::new(move || assert_eq!(*a, [2, 3, 5, 7]));
         d.call();
     }
+
+    #[test]
+    fn long_slice_usize() {
+        let a: [usize; 5] = [2, 3, 5, 7, 11];
+        let mut d = Deferred::new(move || assert_eq!(a, [2, 3, 5, 7, 11]));
+        d.call();
+    }
 }
 ```
+
 </details>
 
 # Drawbacks
--- a/text/2017-08-30-simplify-defer.md
+++ b/text/2017-08-30-simplify-defer.md
@@ -1,6 +1,6 @@
 # Summary
 
-Simplify deferred destruction by removing `Scope::defer_free` and `Scope::defer_drop`.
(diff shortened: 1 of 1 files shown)


arthurprs:
Thanks for the writeup.

I'm not sure I share the sentiment. My impression is that it's not a net-gain as it'll make the most frequent use cases less ergonomic.

The extra Ptr/Owned functions are probably useful on their own, to avoid dirty transmutes and all that.

ghost:
@arthurprs 

Do you at least agree about removing `defer_free`, since we have `ManuallyDrop` now? I thought that was an easy one to sell. :)

> My impression is that it's not a net-gain as it'll make the most frequent use cases less ergonomic.

If we leave `defer_drop` alone, would the RFC then be a net gain for you? -- I'd be okay with that.

text/2017-08-30-simplify-defer.md
```diff
@@ -0,0 +1,182 @@
...
+    Err(h) => head = h,
+}
+```
+
+Neither `Atomic` nor `ManuallyDrop` have a destructor, so dropping a `Node` will be a noop,
+which makes `defer_drop` in this case equivalent to `defer_free`.
+This makes `defer_free` unnecessary.
+This is the kind of problem `ManuallyDrop` was created to solve.
+
+### Do we need `defer_free`?
```
jeehoonkang: maybe `defer_drop`?

text/2017-08-30-simplify-defer.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+Calling the destructor will now come with some performance cost due to additional boxing
+(`defer` stores `FnOnce()` in a `Box`).
+Moreover, deferring destruction this way is a bit more unergonomic.
+Let's fix both problems...
+
+# Detailed design
+
+### `Deferred`
```
jeehoonkang: I think the title here doesn't match with the content.

text/2017-08-30-simplify-defer.md
```diff
@@ -0,0 +1,182 @@
...
+    fn defer<F: FnOnce() + Send + 'static>(&self, f: F);
+
+    // ...
+}
+```
+
+In most cases `defer` will be used to just deallocate memory or execute simple destruction
+routines. The passed `FnOnce()` closure will almost always fit into, say, 3 words.
+
+I have an [implementation](https://gist.github.com/stjepang/00d63b8febc07b297eae3480e80d0e91)
```
jeehoonkang: I wonder if it can be inlined here so that this RFC is self-contained.

text/2017-08-30-simplify-defer.md
```diff
@@ -0,0 +1,182 @@
...
+of `Deferred` ready, which is able to store a small closure within itself, falling back
+to heap allocation for big closures.
+
+**TL;DR:** `Deferred` is to `FnOnce() + Send + 'static` what `SmallVec` is to `Vec`.
```
jeehoonkang: Very helpful analogy :)

text/2017-08-30-simplify-defer.md
```diff
@@ -0,0 +1,182 @@
...
+        scope.defer(move || drop(head));
+        return Some(ptr::read(&h.value));
+    },
+    Err(h) => head = h,
+}
+```
+
+Strictly speaking, this code is possibly incorrect because an `Owned` and `Ptr` pointing to the
+same thing exist at the same time. This is okay because the `Owned` will not be used until the
+closure executes, but **@RalfJung**'s new unsafe code checker might reject the code as unsound.
```
jeehoonkang, line 344: I feel the proposed solution is hacky for the reason that the meaning of `Owned` or `Ptr<'static>` is unclear.  I think it would be great if the meaning can be explained in [RustBelt](http://www.mpi-sws.org/~dreyer/papers/rustbelt/paper.pdf)'s lifetime logic.

Also, I doubt that the `defer*` family of functions is safe.  For example, the client should guarantee that the `defer_drop()`ed object is no longer pointed to by a memory cell.  This knowledge cannot be expressed in Rust's type system, so it should be `unsafe`.

text/2017-08-30-simplify-defer.md
```diff
@@ -0,0 +1,182 @@
...
+    Err(h) => head = h,
+}
+```
+
+# Drawbacks
+
+We'll introduce two new functions (`Owned::from_ptr` and `Ptr::to_static`). But at the same time
+we'll also remove two functions (`Scope::defer_free` and `Scope::defer_drop`).
+
+There might be very small differences in performance when using `defer_free` and `defer_drop` versus
```
jeehoonkang: I think @schets may have an opinion on the performance.  cf: https://github.com/crossbeam-rs/crossbeam-epoch/pull/3#discussion_r129333625

arthurprs: Good point.

arthurprs:
Is it reasonable to assume that defer_drop is the most frequent use case? I can get behind removing defer_free.

text/2017-08-30-simplify-defer.md
ghost: Regarding conversion from `Ptr<'scope>` to `Ptr<'static>`, that is definitely hacky. But I think what we *actually* want here is conversion to `Ptr<'unsafe>`. Unsafe lifetime was proposed in [this RFC](https://github.com/rust-lang/rfcs/pull/1918), but in the end it was postponed.

Regarding safety, `defer_free` and `defer_drop` are definitely unsafe and this RFC doesn't aim to change that. However, `defer` is currently unsafe, but this RFC proposes to make it a safe function. I suppose you're okay with that?

jeehoonkang: Ah, yes.  I am okay with `pub safe fn defer()`.

ghost:
I reduced the scope of the RFC. Now it only proposes to remove `defer_free` and optimize `defer`.

arthurprs (approved):
I'm gonna go ahead and approve it. I'll defer more specific implementation details to the implementation PR.

Vtec234:
I tacked `Deferred` onto the `Garbage` implementation [here](https://github.com/Vtec234/crossbeam-epoch/tree/defer) and then ran [these benchmarks](https://gist.github.com/Vtec234/44690ee4b20780ebfe310eccc33d3828) on it to compare it with the previous implementation (`EPOCH`). The results are as follows:
```
EPOCH:
test bench::bench_defer       ... bench:         487 ns/iter (+/- 19)
test bench::bench_defer_multi ... bench:     139,416 ns/iter (+/- 12,116)
test bench::bench_defer_n     ... bench:      20,005 ns/iter (+/- 314)
test bench::bench_drop        ... bench:         518 ns/iter (+/- 19)
test bench::bench_drop_multi  ... bench:     128,332 ns/iter (+/- 10,473)
test bench::bench_drop_n      ... bench:      42,413 ns/iter (+/- 727)
test bench::bench_free        ... bench:         514 ns/iter (+/- 10)
test bench::bench_free_multi  ... bench:     127,119 ns/iter (+/- 11,145)
test bench::bench_free_n      ... bench:      40,721 ns/iter (+/- 1,411)

DEFERRED:
test bench::bench_defer       ... bench:         687 ns/iter (+/- 14)
test bench::bench_defer_multi ... bench:     186,959 ns/iter (+/- 17,204)
test bench::bench_defer_n     ... bench:      20,119 ns/iter (+/- 1,056)
test bench::bench_drop        ... bench:         692 ns/iter (+/- 13)
test bench::bench_drop_multi  ... bench:     193,229 ns/iter (+/- 15,993)
test bench::bench_drop_n      ... bench:      47,453 ns/iter (+/- 1,640)
test bench::bench_free        ... bench:         688 ns/iter (+/- 18)
test bench::bench_free_multi  ... bench:     192,716 ns/iter (+/- 13,700)
test bench::bench_free_n      ... bench:      45,798 ns/iter (+/- 2,760)
```

Unfortunately, it seems that in threaded benchmarks this version is noticeably slower, presumably due to slower migration of garbage. I might've made some silly mistake and written a nonoptimal implementation, but if not, we have to investigate the causes of this.

jeehoonkang:
## On `defer_free`

My first impression is that `defer_free` might be faster for some cases.  However, now I think `defer_free` is a strange method since we usually don't want to free an object without dropping it.  It would be good to have `defer_drop` only.

For performance, we may need to find a way to determine whether a given type has a trivial drop impl (cf https://github.com/crossbeam-rs/crossbeam/pull/57#issuecomment-277965358).  But we don't have a stable way to do it.  

I agree with the proposed change, and want Rust to stabilize `fn needs_drop()`.

## On `Deferred`

I like the idea, but want to see the benchmark result.  @Vtec234's result looks bad...

Vtec234:
Regarding the API itself, a key difference between `defer_free` and a `defer_drop` with a trivial destructor is that `defer_drop` unnecessarily (but only in the case of trivial drop!) requires `Send + 'static`. Hence, this code:
```rust
use crossbeam_epoch::*;

struct UnSend {
    p: *mut u8,
}

let a = Owned::new(UnSend { p: 0 as *mut u8 } );
pin(|scope: &Scope| {
    unsafe {
        scope.defer_drop(a.into_ptr(scope));
    }
} );
```
will not compile, because pointers are not `Send` - even though we only want to send it for memory freeing purposes, which is fine. Hence, removing `defer_free` without relaxing the requirements on `defer_drop` will suddenly make a class of operations impossible. And of course, relaxing the requirements is a problem we currently have no good solution to.

Vtec234:
I changed my mind and removed a previous comment.

With regards to the API, we have to notice that removing `defer_free` imposes the requirement of `T: Send + 'static` on all objects to be removed, even those with trivial destructors, while `defer_free` allowed recollection of objects without this restriction. I think the `Send` part is perfectly fine - all types used in threaded contexts should be `Send` anyway.

The `'static` part is slightly more troublesome, and a problem we currently have no satisfactory solution to. For now I can't see a usecase where one would have a trivially destructible non-`'static` type and where this change would prevent the usecase from being implemented. However, if one exists, this change would have to be reconsidered.

As the RFC mentions, `defer_drop` could also be removed in favour of `defer`, but that would require adding a few unsafe conversion methods to satisfy trait requirements. This might be a good solution, since it makes explicit the invariants of lifetime that the user has to guarantee, but on the other hand it does make the API slightly more cumbersome to use.

In general, I'm in favour of this RFC, but only if we resolve the performance problems I mentioned in the comment above.

ghost:
@Vtec234

Thank you for writing these benchmarks! I can confirm the performance issues.

It turns out that the size of `Deferred` (as in `std::mem::size_of`) is larger than the size of `Garbage`, which makes it a bit slower. After tweaking the layout of `Deferred` a bit to make it smaller, the issue is resolved:

```
EPOCH
test bench::bench_defer       ... bench:         474 ns/iter (+/- 28)
test bench::bench_defer_multi ... bench:     132,339 ns/iter (+/- 56,620)
test bench::bench_defer_n     ... bench:      21,147 ns/iter (+/- 2,083)
test bench::bench_drop        ... bench:         479 ns/iter (+/- 14)
test bench::bench_drop_multi  ... bench:     124,968 ns/iter (+/- 77,483)
test bench::bench_drop_n      ... bench:      39,859 ns/iter (+/- 4,453)
test bench::bench_free        ... bench:         489 ns/iter (+/- 16)
test bench::bench_free_multi  ... bench:     117,083 ns/iter (+/- 34,990)
test bench::bench_free_n      ... bench:      38,173 ns/iter (+/- 1,925)

DEFERRED
test bench::bench_defer       ... bench:         476 ns/iter (+/- 10)
test bench::bench_defer_multi ... bench:     118,214 ns/iter (+/- 36,568)
test bench::bench_defer_n     ... bench:      16,636 ns/iter (+/- 2,985)
test bench::bench_drop        ... bench:         534 ns/iter (+/- 128)
test bench::bench_drop_multi  ... bench:     114,980 ns/iter (+/- 81,091)
test bench::bench_drop_n      ... bench:      40,418 ns/iter (+/- 5,847)
test bench::bench_free        ... bench:         496 ns/iter (+/- 13)
test bench::bench_free_multi  ... bench:     115,312 ns/iter (+/- 86,908)
test bench::bench_free_n      ... bench:      37,695 ns/iter (+/- 3,172)
```

I've pushed the updated implementation of `Deferred` to this RFC.

text/2017-08-30-simplify-defer.md
```diff
@@ -0,0 +1,365 @@
...
+}
+```
+
+Neither `Atomic` nor `ManuallyDrop` have a destructor, so dropping a `Node` will be a noop,
+which makes `defer_drop` in this case equivalent to `defer_free`.
+This makes `defer_free` unnecessary.
+This is the kind of problem `ManuallyDrop` was created to solve.
+
+# Can we make `defer` safe and faster?
+
```
jeehoonkang, line 80: Maybe three sharps?

jeehoonkang (approved):

Vtec234 (approved):
`After tweaking the layout of Deferred a bit to make it smaller, the issue is resolved`

That's great! If it was this simple to improve performance, surely it will also be possible to tweak it to specific hardware. I think it's fine if we leave the inline implementation of `Deferred` in this RFC as a working PoC and then modify it further in the actual crate code.

ghost:
> The `'static` part is slightly more troublesome, and a problem we currently have no satisfactory solution to. For now I can't see a usecase where one would have a trivially destructible non-`'static` type and where this change would prevent the usecase from being implemented. However, if one exists, this change would have to be reconsidered.

This is a good point and a real concern. I'll open a new issue to discuss how to deal with those lifetime constraints.

PR #10, opened 2017-09-04
Add the generalized-scope RFC

jeehoonkang:
I am proposing:

- making `Scope` trait, and implements `Scope` for the old `Scope` struct.
- letting `unprotected` use a different implementation of `Scope` so that `defer_drop`ped and `defer_free`d objects are immediately dropped and freed.
- supporting multiple EBR instances.

This RFC is fully implemented in [this branch](https://github.com/jeehoonkang/crossbeam-epoch/tree/unprotected).

[Rendered](https://github.com/jeehoonkang/crossbeam-rfcs/blob/generalized-scope/text/2017-09-04-generalized-scope.md)

### DIFF
--- a/text/2017-09-04-generalized-scope.md
+++ b/text/2017-09-04-generalized-scope.md
@@ -8,8 +8,9 @@ region of shared memory.
 # Motivation
 
 The `unprotected` function provides a `Scope` without pinning the thread, where the caller thread is
-the only one accessing atomics. Assuming this, we can **immediately** deallocate and drop objects in
-an unprotected scope. However, even in an unprotected scope, the `defer_free` and `defer_drop`
+either (1) the only one accessing atomics, or (2) no threads are modifying atomics (think:
+reader-writer locks). Assuming this, we can **immediately** deallocate and drop objects in an
+unprotected scope. However, even in an unprotected scope, the `defer_free` and `defer_drop`
 functions defers freeing and dropping the objects until the scope is over. Even worse, if a lot of
 pieces of garbage are created, they are moved to the global garbage queue.
 
--- a/text/2017-09-04-generalized-scope.md
+++ b/text/2017-09-04-generalized-scope.md
@@ -1,0 +1,125 @@
+# Summary
+
+This is a proposal for generalizing the `Scope` struct. It primarily aims to optimize the
+`unprotected` function. As a side benefit, we can easily support multiple instances of EBR-managed
+region of shared memory.
+
+
+# Motivation
+
+The `unprotected` function provides a `Scope` without pinning the thread, where the caller thread is
+the only one accessing atomics. Assuming this, we can **immediately** deallocate and drop objects in
+an unprotected scope. However, even in an unprotected scope, the `defer_free` and `defer_drop`
+functions defers freeing and dropping the objects until the scope is over. Even worse, if a lot of
+pieces of garbage are created, they are moved to the global garbage queue.
+
+Another motivation is the fact that currently we are supporting only a single instance of
+EBR-managed region of shared memory. For example, a bad-behaving mutator can pin itself
+indefinitely so that the global epoch cannot be advanced too long. In order to mitigate this
+problem, we would like to support multiple instances of EBR so that a mutator in an instance cannot
+bother the advancement of another instance's epoch.
+
+
+# Detailed Design
+
+This proposal is fully implemented
+in [this branch](https://github.com/jeehoonkang/crossbeam-epoch/tree/unprotected).
+
+
+## Generalizing `Scope`
+
+In order to enhance the `unprotected` function, we propose:
+
+- Creating
+  the
+  [`Scope` trait](https://github.com/jeehoonkang/crossbeam-epoch/blob/unprotected/src/realm.rs#L15),
+  which provides the methods `Scope` currently has;
+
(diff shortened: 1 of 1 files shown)


text/2017-09-04-generalized-scope.md
```diff
@@ -0,0 +1,125 @@
+# Summary
+
+This is a proposal for generalizing the `Scope` struct. It primarily aims to optimize the
+`unprotected` function. As a side benefit, we can easily support multiple instances of EBR-managed
+region of shared memory.
+
+
+# Motivation
+
+The `unprotected` function provides a `Scope` without pinning the thread, where the caller thread is
```
ghost, line 10: More precisely (see the [Atomic API](https://github.com/crossbeam-rs/rfcs/blob/master/text/2017-05-02-atomic-api.md) RFC), the caller thread is either the only one accessing atomics or no one is modifying atomics. This is sort of like a `RwLock`.

text/2017-09-04-generalized-scope.md
```diff
@@ -0,0 +1,125 @@
...
+the only one accessing atomics. Assuming this, we can **immediately** deallocate and drop objects in
+an unprotected scope. However, even in an unprotected scope, the `defer_free` and `defer_drop`
+functions defers freeing and dropping the objects until the scope is over. Even worse, if a lot of
```
ghost: Would it be a good idea to simply remove `defer_free` and `defer_drop` from `UnprotectedScope`?

Why would someone want to call `defer_free` or `defer_drop` in an unprotected scope?

text/2017-09-04-generalized-scope.md
```diff
@@ -0,0 +1,125 @@
...
+in [this branch](https://github.com/jeehoonkang/crossbeam-epoch/tree/unprotected).
+
+
+## Generalizing `Scope`
+
+In order to enhance the `unprotected` function, we propose:
+
+- Creating
+  the
+  [`Scope` trait](https://github.com/jeehoonkang/crossbeam-epoch/blob/unprotected/src/realm.rs#L15),
```
ghost, line 36: Instead of creating a split between `EpochScope` and `UnprotectedScope`, why not just put `is_unprotected: bool` inside `Scope`?

ghost:
Reviewing only the changes related to `unprotected` now, and will look into realms later.

First of all, I think it's reasonable to optimize code whenever we can. But in this case I'm wondering if this is the right direction to push optimizations into.

Why would someone want to call `defer{,_free,_drop}` in an unprotected scope? I would argue that, if you're calling `defer` inside an unprotected scope, there's a good chance you might be doing something wrong, or inadvertently writing code that does something you don't want to do. It's just such a fishy thing to do. :)

Is the intention to immediately destroy the `Ptr`? After all, manually doing `drop(Box::from_raw(p.as_raw() as *mut Node<>T))` is much uglier than `scope.defer_drop(p)`, so I can see the appeal behind that.

If that is indeed the case, perhaps we'd like an unsafe method `Ptr::destroy` that drops and deallocates the object. Then you could simply do `p.destroy()`. And to follow up a bit, perhaps we could then rename `defer_drop` to `defer_destroy` and implement it like this:

```rust
impl<'scope> Scope<'scope> {
    unsafe fn defer_destroy<T: 'static>(&self, ptr: Ptr<'scope, T>) {
        let ptr: Ptr<'static, T> = mem::transmute(ptr);
        self.defer(move || ptr.destroy());
    }

    // ...
}
```

jeehoonkang:
@stjepang thank you for the feedback!  I would like to answer some of your questions.

On `defer*` methods in unprotected scopes, I initially thought the same thing, but in implementing it in `crossbeam-epoch` I met the following code snippet: https://github.com/crossbeam-rs/crossbeam-epoch/blob/master/src/sync/queue.rs#L184  Here, a queue's drop method repeatedly `try_pop()` its elements in an unprotected scope.  Which is justified because the calling thread is the only one accessing the queue.  However, inside `try_pop()`, a node of the queue is `defer_drop()`ped.  That's the primary reason I implemented `defer_*` for unprotected scopes.

Unprotected scopes do not need a garbage bag.  So using the `is_protected` boolean field in `Scope` instead of using `UnprotectedScope` will incur performance cost, which is probably small, though.

I very much like the idea of `Ptr::destroy()`.  If there is no significant regressions, I would like to go for it.

ghost:
> I initially thought the same thing, but in implementing it in `crossbeam-epoch` I met the following code snippet

I see -- this is indeed a good case for calling `defer*` inside an unprotected scope.

> Unprotected scopes do not need a garbage bag. So using the `is_protected` boolean field in `Scope` instead of using `UnprotectedScope` will incur performance cost, which is probably small, though.

Actually, we don't need an additional boolean field. [Here](https://github.com/crossbeam-rs/crossbeam-epoch/blob/7fc2813205c53ef7924ee3e0895e258055f79ccb/src/mutator.rs#L234) we should just check whether `self.bag` is null. If it is null, we destroy garbage immediately, otherwise we push it into the bag.

This is a very easy solution, and now there's no need to generalize scopes. :) 
The only thing left to do is to clarify in the documentation what `defer*` does in unprotected scopes.

jeehoonkang:
I'm closing it.  In #12, we are discussing what will be the right interface for the "realms", which might evolve to an RFC later.

PR #15, opened 2017-10-07
Introduce `Collector` API

jeehoonkang:
This is an RFC for introducing the `Collector` API.

Blocks https://github.com/crossbeam-rs/crossbeam-epoch/pull/21

[Rendered](https://github.com/jeehoonkang/crossbeam-rfcs/blob/api-layers/text/2017-10-06-collector-api.md)

### DIFF
--- a/text/2017-10-06-api-layers.md
+++ b/text/2017-10-06-api-layers.md
@@ -57,11 +57,13 @@ impl Global {
 impl Local {
     fn new(global: &Global) -> Self;
 
-    pub fn pin<F, R>(&self, global: &Global, f: F)
+    // `global` should be the one used to create the `Local`.
+    pub unsafe fn pin<F, R>(&self, global: &Global, f: F)
     where F: for<'scope> FnOnce(Scope<'scope>) -> R;
     // fun f with a `Scope`
 
-    pub unsafe fn finalize(&self, global: &Global);
+    // Once a `Local` is unregistered, `pin()` or `unregister()` should not be called later.
+    pub unsafe fn unregister(&self, global: &Global);
     // deregisters itself from the global.
 
     ... // other methods
@@ -169,7 +171,7 @@ impl Deref for Handle {
 
 impl Drop for Handle {
     fn drop(&mut self) {
-        self.finalize(&GLOBAL);
+        unsafe { self.unregister(&GLOBAL); }
     }
 }
 
--- a/text/2017-10-06-api-layers.md
+++ b/text/2017-10-06-api-layers.md
@@ -1,0 +1,182 @@
+# Summary
+
+This RFC introduces three layers of API for supporting diversified use cases.
+
+
+# Motivation
+
+As the Crossbeam project is growing, users want more diversified API than the existing one.
+
+First, majority of users will be happy with the current API, which interacts with the default
+singleton garbage collector.
+
+Second, some users may want to create their own garbage collector, and use the collector via a
+*handle* to it. Since the current API provides only one garbage collector, Crossbeam currently does
+not support this use case.
+
+Lastly, some users want to embed a garbage collector in another systems library, e.g. memory
+allocator and thread management environment.  Note that it is often a part of the standard library,
+thus this use case requires (at least a part of) Crossbeam to be implemented in the `#[no_std]`
+environment. Likewise, Crossbeam currently does not support this use case.
+
+In order to support these diversified use cases, this author proposes to generalize the API by
+introducing an API layer for each use case: (1) **the default garbage collector API** for the first
+use case; (2) **`Collector` API** for the second one; and (3) **internal API** for the third one.
+
+
+
+# Detailed design
+
+FYI, this RFC is fully implemented in [this
+branch](https://github.com/jeehoonkang/crossbeam-epoch/tree/handle). You can find a discussion on
+this branch in [this PR](https://github.com/crossbeam-rs/crossbeam-epoch/pull/21).
+
+
+## Internal API
+
+This is the lowest-level API to Crossbeam. First, we define a public struct `Global` and
+`Participant` that contain the global and local data for garbage collection:
+
+```rust
+struct Global {
+  ...
+}
+
+struct Participant {
+  ...
+}
+```
+
+And implement the following methods:
+
+```rust
+impl Global {
+    pub fn new() -> Self;
+}
+
+impl Participant {
+    fn new(global: &Global) -> Self;
+
+    pub fn pin<F, R>(&self, global: &Global, f: F)
+    where F: for<'scope> FnOnce(Scope<'scope>) -> R;
+    // fun f with a `Scope`
+
+    ... // other methods
+}
+```
+
+Note that `Participant::pin()` requires a reference to a `Global`.
+
+When you want to embed a garbage collector in another systems library, e.g. logger, you first define
+its own `Global` and `Participant` structs, and embed `epoch::Global` and `epoch::Participant` in
+corresponding structs, as follows:
+
+```rust
+pub mod Logger {
+
+pub struct Global {
+    epoch: epoch::Global,
+    ... // more global data
+}
+
+pub struct Participant {
+    epoch: epoch::Participant,
+    ... // more local data
+}
+
+...
+
+}
+```
+
+You can easily finish the internal logger API by implementing the methods of `Global` and
+`Participant` using `epoch`. You can build the `Logger` API and the default logger API in the same
+way as below.
+
+For a full example, see [this repository](https://github.com/jeehoonkang/handle-example-rs).
+
+
+## `Collector` API
+
+The public struct `Collector` is a general-purpose garbage collector with its own global epoch, and
+the public struct `Handle` is an abstraction of a participant in a collector.  `Collector` is a
+counted reference to a `Global`, and `Handle` is a tuple of a counted reference to a `Global` and a
+`Participant`:
+
+```rust
+struct Collector(Arc<Global>);
+
+struct Handle {
+    global: Arc<Global>,
+    local: Local,
+}
+```
+
+You can create a new collector by creating a new `Global` data, and add a handle to a collector by
+sharing its `Global` data.  Using a `Handle`, you can create a `Scope` and pass it to a given
+function:
+
+```rust
+impl Collector {
+    pub fn new() -> Self;
+    pub fn handle(&self) -> Handle;
+}
+
+impl Handle {
+    pub fn pin<F, R>(&self, f: F) 
+    where F: for<'scope> FnOnce(Scope<'scope>) -> R {
+        self.participant.pin(&self.global, f)
+    }
+
+    ... // other methods
+}
+```
+
+Note that `Handle::pin()`, on contrary to `Participant::pin()`, does not require a reference to a
+`Global`, because `Handle` already has a reference.
+
+See [the
+implementation](https://github.com/jeehoonkang/crossbeam-epoch/blob/handle/src/collector.rs) for
+more details.
+
+
+## The default garbage collector API
+
+There will be the default collector and per-thread handle to the default collector:
+
+```rust
+lazy_static! { pub GLOBAL: Global = Global::new(); }
+thread_local! { pub PARTICIPANT: Participant = Participant::new(&GLOBAL); }
+
+pub fn pin<F, R>(f: F) 
+where F: for<'scope> FnOnce(Scope<'scope>) -> R {
+    PARTICIPANT.with(|participant| { participant.pin(&GLOBAL, f) })
+}
+```
+
+See [the implementation](https://github.com/jeehoonkang/crossbeam-epoch/blob/handle/src/default.rs)
+for more details.
+
+
+
+# Alternatives
+
+`Scope<'scope>` vs. `&'scope Scope`: Currently the type of scope is `Scope<'scope>`. It is possible
+to use more traditional reference type `&'scope Scope`, but it may incur some runtime cost.
+
+`Arc<Global>` vs. `&Global`: Currently `Collector` is implemented as `Arc<Global>`, which incurs
+runtime overhead of reference counting.  Using `&Global` will eliminate the runtime cost, but it
+will significantly complicates the API.
+
+
+
+# Unresolved questions
+
+For unknown reasons, there is a significant performance regression in the `pin_holds_advance()` test
+function.
+
+for `#[no_std]` environment, a long-term plan would be separating out what is dependent on `std`,
+namely the default collector API and the `Collector` API, and what is not, namely the internal
+API. The proposed internal API is almost `std`-free except for the fact that `Global` and `Local`
+relies on the data structures in the standard library. This author believes we can finish this job
+once [the allocator trait is stabilized](https://github.com/rust-lang/rust/issues/32838).
--- a/text/2017-10-06-api-layers.md
+++ b/text/2017-10-06-api-layers.md
@@ -64,7 +64,7 @@ impl Local {
 
     // Once a `Local` is unregistered, `pin()` or `unregister()` should not be called later.
     pub unsafe fn unregister(&self, global: &Global);
-    // deregisters itself from the global.
+    // unregisters itself from the global.
 
     ... // other methods
 }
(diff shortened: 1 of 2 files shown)


text/2017-10-06-api-layers.md
```diff
@@ -0,0 +1,182 @@
...
+
+
+
+# Alternatives
+
+`Scope<'scope>` vs. `&'scope Scope`: Currently the type of scope is `Scope<'scope>`. It is possible
+to use more traditional reference type `&'scope Scope`, but it may incur some runtime cost.
+
+`Arc<Global>` vs. `&Global`: Currently `Collector` is implemented as `Arc<Global>`, which incurs
+runtime overhead of reference counting.  Using `&Global` will eliminate the runtime cost, but it
```
arthurprs: Is this a problem in practice? As far as I can see Arc::clone(..) happens when creating Collectors and Handles. The first is very likely a non issue. The second I'm not sure, but I guess creating Handles won't be in people hot path anyway.

jeehoonkang: Thank you for the comment!  I slightly edited the text, articulating the fact that we chose `Arc<Global>` over `&Global` because of exactly what you said.

text/2017-10-06-api-layers.md
```diff
@@ -0,0 +1,206 @@
...
+not support this use case.
+
+Lastly, some users want to embed a garbage collector in another systems library, e.g. memory
+allocator and thread management environment.  Note that it is often a part of the standard library,
+thus this use case requires (at least a part of) Crossbeam to be implemented in the `#[no_std]`
+environment. Likewise, Crossbeam currently does not support this use case.
+
+In order to support these diversified use cases, this author proposes to generalize the API by
+introducing an API layer for each use case: (1) **the default garbage collector API** for the first
+use case; (2) **`Collector` API** for the second one; and (3) **internal API** for the third one.
```
ghost: Why is `Collector` API not suitable for the third use case (e.g. things like elfmalloc), but internal API is?

What is the crucial difference between them? Is it that `Collector` API is supposed to be used in `std` environment and internal API in `#[no_std]` environment?

text/2017-10-06-api-layers.md
```diff
@@ -0,0 +1,206 @@
...
+
+```rust
+impl Global {
+    pub fn new() -> Self;
+}
+
+impl Local {
+    fn new(global: &Global) -> Self;
+
+    pub fn pin<F, R>(&self, global: &Global, f: F)
```
ghost: What happens if we pass an incorrect reference to `Local::pin` or `Local::finalize`? Is that safe?

text/2017-10-06-api-layers.md
```diff
@@ -0,0 +1,206 @@
...
+    where F: for<'scope> FnOnce(Scope<'scope>) -> R;
+    // fun f with a `Scope`
+
+    pub unsafe fn finalize(&self, global: &Global);
```
ghost: Nit: `unregister` might fit better than `finalize`

text/2017-10-06-api-layers.md
```diff
@@ -0,0 +1,206 @@
...
+See [the implementation](https://github.com/jeehoonkang/crossbeam-epoch/blob/handle/src/default.rs)
+for more details.
+
+
+
+# Alternatives
+
+`Scope<'scope>` vs. `&'scope Scope`: Currently the type of scope is `Scope<'scope>`. It is possible
+to use more traditional reference type `&'scope Scope`, but it may incur some runtime cost of double
+indirection.
```
ghost: Well, in your pull request, `Scope` currently looks like this:
```rust
pub struct Scope<'scope> {
    /// A reference to the global data.
    pub(crate) global: *const Global,
    /// A reference to the thread-local bag.
    pub(crate) bag: *mut Bag, // !Send + !Sync
    /// `global` is effectivey `&'scope Global`
    pub(crate) _marker: PhantomData<&'scope Global>,
}
```

This `_marker` is not really needed - you can just delete it. :) And then we can keep using traditional references: `&'scope Scope`.

text/2017-10-06-api-layers.md
```diff
@@ -0,0 +1,206 @@
...
+
+
+
+# Unresolved questions
+
+for `#[no_std]` environment, a long-term plan would be separating out what is dependent on `std`,
+namely the default collector API and the `Collector` API, and what is not, namely the internal
+API. The proposed internal API is almost `std`-free except for the fact that `Global` and `Local`
+relies on the data structures in the standard library. This author believes we can finish this job
+once [the allocator trait is stabilized](https://github.com/rust-lang/rust/issues/32838).
```
ghost: To support `#[no_std]` environments, do we want:

1. An API that doesn't allocate memory at all?
2. An API that allocates memory, but through the `Allocator` trait?

I'm guessing the answer is (2). But if that is the case, then it's trivial to implement `Arc` that uses the `Allocator` trait. So we could still easily have the `Collector` API in `#[no_std]`, right?

jeehoonkang: I agree that we can easily support the `Collector` API in `#[no_std]`.  I changed the text.

That makes me think more of the organization of the Crossbeam crates.  I'd like to suggest we put the internal API (if we decide to expose it) and the `Collector` API in `crossbeam-epoch`, and implement the default garbage collector in the `crossbeam` crate.  A long-term goal is implementing `crossbeam-epoch` in the `#[no_std]` environment.  And those who just want to the default collector will import `crossbeam`, and those who want to use their own collector will import `crossbeam-epoch`.

text/2017-10-06-api-layers.md
jeehoonkang: Yes, you can use the `Collector` API for e.g. `elfmalloc`.  In particular, as you pointed out, `Collector` API can be easily adjusted for being used in `#[no_std]` environment.  I think the main difference between the internal and the `Collector` APIs is the number of pointer indirection.


Using the `Collector` API, `elfmalloc`'s `Global` may look like:

```rust
struct Global {
    collector: epoch::Collector,
    ... // other global data
}
```

Note that `epoch::Collector` is just `Arc<epoch::Global>` under the hood.  So there is an additional pointer indirection for accessing the global data for the underlying collector from a `elfmalloc` handle.  The situation gets worse if you want to embed `elfmalloc` in even another systems library: two pointer indirections are required for accessing the global data for the collector from the enclosing library's handle.


If you use the internal API, there is no additional indirection for accessing the global data for the collector, as evident in the following hypothetical definition of `elfmalloc`'s `Global`:

```rust
struct Global {
    epoch: epoch::Global,
    ... // other global data
}
```

As I said in the RFC text, `elfmalloc` (or a logger) may expose their own the `Allocator` (or `Logger`) API, and the default allocator (or logger) API.


I'd like to also ask @joshlf's opinion.  He is a main developer of `elfmalloc`.

arthurprs:
@jeehoonkang I see that you removed the line about perf. regression, is it solved?

text/2017-10-06-api-layers.md
ghost: Interesting idea.

Another thing we should consider (or at least think about) is doing something akin to Rayon's split into `rayon` and `rayon-core`.

For example, let's say that we bump the version of `crossbeam` from `0.4` to `0.5` because we've just added a new data structure to it. Both versions depend on `crossbeam-epoch="0.1"`. Now, suppose that you have your own project that depends on crates A and B. Crate A depends on `crossbeam="0.4"` and crate B depends on `crossbeam="0.5"`. It'd be great if they both used the same global `Collector`.

This makes me think that the global `Collector` shouldn't live inside `crossbeam`, but rather inside a crate that increments versions very slowly (ideally, always at version `1.0.x`). That would be `crossbeam-epoch`, or perhaps even `crossbeam-epoch-global`. This is similar to `rayon-core`.

I'll have to think about all this a bit more. :) Perhaps we should open a separate issue to discuss.

text/2017-10-06-api-layers.md
ghost: Hmm. I agree that there is one level of indirection less, but what kind of operations would actually benefit from this?

Suppose we're pinning locals/handles.
In order to pin a `Local`, you'd do: `my_local.pin(&my_global, || ...)`.
But if using `Collector` API, then you have: `my_handle.pin(|| ...)`.
The first case has the manually supplied `&my_global`, while the second one simply reaches for the `Arc<Collector>` inside `my_handle`. Both cases are actually very similar - I'd be very surprised to see any tangible difference in performance here.

But sometimes we might avoid locals/handles and reach for the global/collector directly, for example: `my_global.collect()` or `my_global.has_garbage()`. But even then I don't expect any noticeable difference, and can hardly imagine such method calls being on hot paths in the code anyway. :)

text/2017-10-06-api-layers.md
```diff
@@ -0,0 +1,211 @@
...
+impl Local {
+    fn new(global: &Global) -> Self;
+
+    // `global` should be the one used to create the `Local`.
+    pub unsafe fn pin<F, R>(&self, global: &Global, f: F)
+    where F: FnOnce(&Scope) -> R;
+    // fun f with a `Scope`
+
+    // Once a `Local` is unregistered, `pin()` or `unregister()` should not be called later.
+    pub unsafe fn unregister(&self, global: &Global);
```
joshlf: Is the idea that this is called when you're not going to use the `Local` again? So it's basically a manual `drop` for `Local`?

text/2017-10-06-api-layers.md
```diff
@@ -0,0 +1,211 @@
...
+
+
+# Unresolved questions
+
+For `#[no_std]` environment, a long-term plan would be separating out what is dependent on `std`,
+namely the default collector API, and what is not, namely the `Collector` API and the internal
+API. The proposed `Collector` and internal API is almost `std`-free except for the fact that (1) the
+`Collector` API uses `std::sync::Arc`, and (2) they rely on the data structures in the standard
+library. This author believes we can easily make it fully independent from `std` by (1)
+re-implementing `Arc` in `#[no_std]`, and (2) using a `#[no_std]` data structures after [the
```
joshlf: I'm hoping that `Arc` will eventually be provided in no-std (libcore, libcollections, etc) with an `Alloc` parameter. That's probably a ways away though.

text/2017-10-06-api-layers.md
joshlf: I'll let @ezrosent chime in here since he's done more work on [`bagpipe`](https://crates.io/crates/bagpipe) than I have (which is the code in elfmalloc that actually uses epoch gc), but I think this would work for us.

text/2017-10-06-api-layers.md
ghost: Yes.

text/2017-10-06-api-layers.md
jeehoonkang: @stjepang Thank you for the analysis.  I was handwaving on this matter and didn't think of the performance overhead precisely.  Now I know the tradeoff clearly.

Now I'm fine with removing the internal API for this RFC and the corresponding PR, provided that `elfmalloc` developers are happy with the `Collector` API.

Though I still prefer exposing the internal API for the last bit of optimization.  I think we can discuss this matter later in another RFC after we gather more real-world use cases of Crossbeam.

jeehoonkang:
@arthurprs Yes, I forgot to call `Local::unregister()`.  After calling it, the performance regression is gone.  That said, I have to admit that `Local` is not a user-friendly API: it's easy to forget to call `unregister()`, and forgetting to call `unregister()` may incur unbearable performance overhead.

text/2017-10-06-api-layers.md
joshlf: Same caveat about @ezrosent knowing this better than I do, but I'm pretty sure we'd be fine with the `Collector` API so long as it can eventually support a parametric `Alloc` (which, as I mentioned in a different comment, depends on `Arc` supporting a parametric `Alloc`).

jeehoonkang: I removed the internal API from this RFC and the corresponding PR.  I think there are no more unresolved issues, and they are ready to be merged.  Please give another round of reviews.  Thanks!

text/2017-10-06-collector-api.md
```diff
@@ -0,0 +1,147 @@
...
+
+struct Handle(Local);
+
+impl Handle {
+    fn new() -> Self {
+        Self { 0: Local::new(&GLOBAL) }
+    }
+}
+
+impl Deref for Handle {
```
ghost: We probably don't want automatic dereferencing of public type into a private type.

ghost (approved):
I'm quite happy with the state of this RFC -- approving.

text/2017-10-06-api-layers.md
Vtec234: > In order to pin a Local, you'd do: my_local.pin(&my_global, || ...).
But if using Collector API, then you have: my_handle.pin(|| ...).
The first case has the manually supplied &my_global, while the second one simply reaches for the Arc\<Collector\> inside my_handle.

Just to clarify, the second case reaches for `Arc<Global>` inside the handle, which is also just one level of indirection. If it reached for `Collector` first, there would be two levels.

> I removed the internal API from this RFC and the corresponding PR.

I think it's fine if we don't expose it for now and only make the `Arc`-based API available, possibly adding the internal API later.

text/2017-10-06-collector-api.md
```diff
@@ -0,0 +1,147 @@
...
+
+There will be the default collector and per-thread handle to the default collector:
+
+```rust
+lazy_static! { pub GLOBAL: Global = Global::new(); }
+thread_local! { pub HANDLE: Handle = Handle::new(); }
+
+struct Handle(Local);
+
+impl Handle {
```
Vtec234: I understand that this type is different than the `pub struct Handle` introduced above, since it has different members and is meant for usage in a default context only. For this reason, maybe it should be called something else like `DefaultHandle`? It's a bit confusing if they're named the same.

jeehoonkang: @Vtec234 Thank you for suggestions.   I renamed it to `DefaultHandle`, as @Vtec234 said.  It's actually a private type that is different from the one exposed in the API.

@stjepang I think it is okay to implement `Deref` for `DefaultHandle`, as it is private.

text/2017-10-06-collector-api.md
```diff
@@ -0,0 +1,147 @@
...
+more details.
+
+
+## The default garbage collector API
+
+There will be the default collector and per-thread handle to the default collector:
+
+```rust
+lazy_static! { pub GLOBAL: Global = Global::new(); }
+thread_local! { pub HANDLE: Handle = Handle::new(); }
```
ghost: Do we still need those `Global` and `Handle` types? Why don't we have this:

```rust
lazy_static! { pub COLLECTOR: Collector = Collector::new(); }
thread_local! { pub HANDLE: Handle = COLLECTOR.handle(); }
```

text/2017-10-06-collector-api.md
```diff
@@ -0,0 +1,147 @@
...
+
+impl Deref for DefaultHandle {
+    type Target = Local;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Drop for DefaultHandle {
```
ghost: I believe we don't need this `Drop` implementation anymore.

text/2017-10-06-collector-api.md
Vtec234: The `DefaultHandle` has one less pointer indirection, since it's contained in the lazy_static. With `Arc` you have to get to the static first, and then to `ArcInner`. We could replace them if there is no demonstrable overhead to doing this.

ghost: Note that `lazy_static!` contains a `Once` that does branching and an atomic `SeqCst` load on every access (`deref`). It also [boxes](https://github.com/rust-lang-nursery/lazy-static.rs/blob/master/src/lazy.rs#L23) the value (unless you use nightly) so there's still a pointer indirection. So `lazy_static!` may in fact be even worse.

Vtec234: Hm, I did forget that `Handle` contains an `Arc<Global>` member, so we don't have to reach through the static. In that case I support the idea to completely remove `DefaultHandle` and replace it with the new types.

jeehoonkang: I updated the RFC and the corresponding PR.  Now we use `Collector` for the default collector. 
 Please have a look!

Vtec234 (approved):
LGTM

jeehoonkang:
Thank you for all the comments!  This RFC is tightly coupled with the corresponding PR.  Shall we merge them together?  Or we'll have a review round for the PR?

ghost:
Let's merge this RFC. 🎉🎉🎉

PR #17, opened 2017-10-14
Consider dynamic particiation in the Crossbeam proof

jeehoonkang:
I intentionally left the nomenclature (e.g. `unlink()`) as-is, because I'd like to focus on fixing rather than improving an already accepted RFC.  Maybe we can revisit the proof and write an RFC that supercedes the existing proof, and there we can revise the text as we wish.


ghost (approved):
Looks good to me.

jeehoonkang:
I'm merging it because it's (arguably) simple fix.  Anyway, proof is just a text.. :)

PR #20, opened 2017-11-03
Replace scopes with guards

### LINKED ISSUE #19: Guards can be sound
In the [Atomic API RFC](https://github.com/stjepang/rfcs-crossbeam/blob/atomic-api/text/2017-05-02-atomic-api.md#scopes-or-guards), I argued that we should model pinning using scopes rather than guards, for three reasons:

1. Guards are tricky to implement correctly (in a sound manner).
2. Guards were a tiny bit slower in my microbenchmarks.
3. Guards are less explicit and thus easier to misuse (by e.g. accidentally pinning for too long).

This time I'd just like to show that the first point is invalid - it's actually pretty easy to make guards sound.

Here's a quick refresher. The problem was the following: if you pin the current thread and obtain a guard, you can keep having the guard for a very long time, even after the thread-local `Handle` gets destructed. Using it after that would be unsafe.

And here's the solution... Currently, when traversing the linked list of registered threads, we remove a node if its next pointer is tagged with 1. We should just have an additional check: remove it if the tag is 1 *and* if the thread is unpinned. That's all we need to change.

cc @hniksic Guards would make the following possible (mentioned [here](https://www.reddit.com/r/rust/comments/763o7r/blog_posts_introducing_lockfree_rust_comparing/docu634/)):

```rust
pub get_ref(&'a self, key: K) -> Option<ValueGuard<'a, V>>;
```

ghost:
[Rendered](https://github.com/stjepang/rfcs-crossbeam/blob/guard/text/2017-11-02-guards.md)

[Implementation](https://github.com/crossbeam-rs/crossbeam-epoch/pull/31)

Closes #19.

### DIFF
--- a/text/2017-11-02-guards.md
+++ b/text/2017-11-02-guards.md
@@ -76,8 +76,8 @@ Each `Local` holds a `ManuallyDrop<Arc<Global>>`, thus keeping the garbage colle
 which it resides alive.
 Similarly, each `Handle` and each `Guard` keeps a pointer to the `Local` associated with it.
 Handles and guards are counted using fields `handle_count` and `guard_count` inside `Local`.
-When both counts reach zero, the `Arc<Global>` inside `Local` is immediately dropped
-(using `ManuallyDrop::drop`), and the heap-allocated `Local` is then marked as deleted.
+When both counts reach zero, the heap-allocated `Local` is marked as deleted and then its
+`Arc<Global>` is immediately dropped (using `ManuallyDrop::drop`).
 If that `Local` was holding the last reference to the `Global`, then the
 `Global` is destroyed as well.
 
@@ -180,21 +180,71 @@ but with non-`Send` handles that cloning behavior is not very useful anymore.
 
 ### Accessing the default handle
 
-Moreover, we'been having difficulties finding a satisfying signature for the
+Moreover, we've been having difficulties finding a satisfying signature for the
 `default_handle` method (see [PR #28](https://github.com/crossbeam-rs/crossbeam-epoch/pull/28)).
-The dilemma is whether we should have:
 
-1. Unsafe function `default_handle` returning a `&'static Handle`.
-2. Safe function `with_default_handle` taking a closure (similar to `LocalKey::with`).
-   This is slightly unergonomic.
-3. Safe function `default_handle` that uses reference counting (`Arc`) to return a
-   reference to the default handle. This would be very similar to the behavior of
-   `std::thread::current()`. The main problem is that cloning an `Arc` would be slow.
+Here is a list of options we have to choose from:
 
-But if handles are not `Send`, a fourth option emerges:
+#### 1st option
 
-4. Same as the third option, but have internal reference counting using `Rc` instead of `Arc`.
-   This greatly reduces the incurred performance costs.
+As proposed in [PR #28](https://github.com/crossbeam-rs/crossbeam-epoch/pull/28), and with
+an additional `try_default_handle` function gated under the `nightly` feature
+(because `LocalKey::try_with` is still unstable):
+
+```rust
+pub unsafe fn default_handle() -> &'static Handle {
+    &*HANDLE.with(|handle| handle as *const _)
+}
+
+#[cfg(feature = "nightly")]
+pub unsafe fn try_default_handle() -> Option<&'static Handle> {
+    HANDLE.try_with(|handle| &*(handle as *const _)).ok()
+}
+```
+
+Advantage: Zero-cost.
+
+Drawback: Functions are `unsafe` (because the `'static` lifetime is a lie).
+
+#### 2nd option
+
+These functions are similar to `LocalKey::with` and `LocalKey::try_with` in that
+they take a closure that operates on a `&Handle`:
+
+```rust
+pub fn with_default_handle<R, F: FnOnce(&Handle) -> R>(f: F) -> R {
+    HANDLE.with(|handle| f(handle))
+}
+
+#[cfg(feature = "nightly")]
+pub fn try_with_default_handle<R, F: FnOnce(&Handle) -> R>(f: F) -> Option<R> {
+    HANDLE.try_with(|handle| f(handle)).ok()
+}
+```
+
+Advantage: Zero-cost.
+
+Drawback: Slightly inconvenient interface.
+
+#### 3rd option
+
+If we make handles non-`Send`, now we can count the number of handles inside `Local`
+and provide a very similar interface to `std::thread::current()`:
+
+```rust
+pub fn default_handle() -> Handle {
+    HANDLE.with(|handle| handle.clone())
+}
+
+#[cfg(feature = "nightly")]
+pub fn try_default_handle() -> Option<Handle> {
+    HANDLE.try_with(|handle| handle.clone()).ok()
+}
+```
+
+Advantage: The most convenient interface.
+
+Drawback: There is a small cost associated with incrementing/decrementing the internal handle count.
 
 ### The new interface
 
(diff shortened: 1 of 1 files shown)


text/2017-11-02-guards.md
```diff
@@ -0,0 +1,213 @@
...
+thread_local! {
+    static HANDLE: Handle = COLLECTOR.handle();
+}
+
+pub fn default_handle() -> Handle {
+    HANDLE.with(|handle| handle.clone())
+}
+```
+
+This way `default_handle` bears some cost of reference counting, but it should be
```
jeehoonkang: I'm not sure how this solves the problem discussed in https://github.com/crossbeam-rs/crossbeam-epoch/pull/28.  What's the benefit of this over the version in the PR? (https://github.com/crossbeam-rs/crossbeam-epoch/pull/28/files#diff-aeb58025d21f8e06a38eae37fad74b8dR26)

text/2017-11-02-guards.md
```diff
@@ -0,0 +1,213 @@
...
+forgivable, as is demonstrated by the benchmarks...
```
jeehoonkang: It seems there's 1ns diff between `epoch::pin()` and `epoch::default_handle().pin()` (12ns vs. 13ns).  How about specifically pointing out this benchmark result here?

jeehoonkang (commented):
Thanks for the RFC!  I'm almost approving this RFC, but I'd like to discuss the section on `default_handle()` a little bit more.  I left two comments.

text/2017-11-02-guards.md
ghost: Done.

text/2017-11-02-guards.md
ghost: I've added a more elaborate explanation to the text.
The main benefit is that `default_handle` is a safe function.

text/2017-11-02-guards.md
```diff
@@ -0,0 +1,316 @@
...
+
+For that reason, `Handle` will not be `Send`.
+
+Note that, currently, `Handle::clone` creates a brand new handle that can be passed to
+another thread (see [PR #26](https://github.com/crossbeam-rs/crossbeam-epoch/pull/26)),
+but with non-`Send` handles that cloning behavior is not very useful anymore.
+
+### Accessing the default handle
+
+Moreover, we'been having difficulties finding a satisfying signature for the
```
jeehoonkang: "we've been"

text/2017-11-02-guards.md
```diff
@@ -0,0 +1,316 @@
...
+data associated with it (in struct `Local`). All such `Local`s are connected into a
+linked list and the head pointer of that list is held in the global data (in struct `Global`).
+`Global` also holds the global epoch and the global garbage queue.
+
+Each `Local` holds a `ManuallyDrop<Arc<Global>>`, thus keeping the garbage collector in
+which it resides alive.
+Similarly, each `Handle` and each `Guard` keeps a pointer to the `Local` associated with it.
+Handles and guards are counted using fields `handle_count` and `guard_count` inside `Local`.
+When both counts reach zero, the `Arc<Global>` inside `Local` is immediately dropped
+(using `ManuallyDrop::drop`), and the heap-allocated `Local` is then marked as deleted.
```
jeehoonkang: If dropping `Arc<Global>` happens before marking `Local` as deleted, isn't it possible that `Global` is dropped so that every associated `Local`'s are dropped, after which marking the `Local` as deleted is use-after-free?

text/2017-11-02-guards.md
```diff
@@ -0,0 +1,316 @@
...
+1. Unsafe function `default_handle` returning a `&'static Handle`.
+2. Safe function `with_default_handle` taking a closure (similar to `LocalKey::with`).
+   This is slightly unergonomic.
+3. Safe function `default_handle` that uses reference counting (`Arc`) to return a
+   reference to the default handle. This would be very similar to the behavior of
+   `std::thread::current()`. The main problem is that cloning an `Arc` would be slow.
+
+But if handles are not `Send`, a fourth option emerges:
+
+4. Same as the third option, but have internal reference counting using `Rc` instead of `Arc`.
```
jeehoonkang: I think the options 3. and 4. are as unsafe as 1. and 2., as they may panic: https://doc.rust-lang.org/1.21.0/std/thread/struct.LocalKey.html#panics . For this reason, I think the signature of `thread_local!`'s `HANDLE.with()` is misleading: while the signature says it's safe, it is actually quite unsafe.  (I'd like to cc @joshlf, who met this panic in the development of elfmalloc.)

That being said, I'd like to advocate for option 1., which doesn't incur runtime performance.

text/2017-11-02-guards.md
ghost: Good point!

This is a mistake - it's actually the other way around. First, the `Local` is marked as deleted, and then the `Arc<Global>` is dropped. I'll update the text.

text/2017-11-02-guards.md
ghost: I've rewritten this section of the text. Hopefully it's clearer now and objectively lays out the pros & cons of each option. I've also removed the 3. option.

> For this reason, I think the signature of thread_local!'s HANDLE.with() is misleading: while the signature says it's safe, it is actually quite unsafe. 

Just to be clear: when you say "it's actually quite unsafe", do you mean "it's actually quite dangerous", or are you talking about "unsafe" as in "invokes undefined behavior"?

ghost:
@joshlf Could you please take a look at the [*Accessing the default handle*](https://github.com/stjepang/rfcs-crossbeam/blob/guard/text/2017-11-02-guards.md#accessing-the-default-handle) section, since it is related to elfmalloc? You've been dealing with thread-locals and tuning performance a lot, so your opinion would be very helpful here.

@jeehoonkang is advocating for the 1st option, while I want the 3rd. But in the end, I don't feel particularly strongly about it.

However, there is one thing I do strongly care about: let's avoid unsafe APIs whenever possible. In that spirit, the 2nd option would be better than the 1st one.

text/2017-11-02-guards.md
ghost: Fixed.

ghost:
@jeehoonkang 

I appreciate your concerns about the performance of the 3rd option. However, I'd like put some things into perspective first. Take a look at the following benchmarks:

```rust
#[bench]
fn pin_empty(b: &mut Bencher) {
    b.iter(|| epoch::pin());
}

#[bench]
fn default_handle_pin(b: &mut Bencher) {
    b.iter(|| epoch::default_handle().pin());
}

#[bench]
fn local_handle_pin(b: &mut Bencher) {
    let handle = epoch::default_handle();
    b.iter(|| handle.pin());
}

#[bench]
fn thread_local_handle_pin(b: &mut Bencher) {
    thread_local! {
        static MY: Handle = epoch::default_handle();
    }
    b.iter(|| MY.with(|handle| handle.pin()));
}
```

Results:

```
test default_handle_pin      ... bench:          13 ns/iter (+/- 1)
test local_handle_pin        ... bench:           9 ns/iter (+/- 1)
test pin_empty               ... bench:          12 ns/iter (+/- 1)
test thread_local_handle_pin ... bench:           9 ns/iter (+/- 0)
```

Looks like pinning itself costs 9 nanoseconds. Accessing the thread-local (that's `LOCAL.with(...)` inside `epoch::pin()`) adds 3 more nanoseconds. Finally, incrementing/decrementing the reference count to access the default handle costs 1 more nanosecond.

However, if you store the handle in an another thread-local (in this example, `MY`) and access it with `MY.with(...)`, you don't pay the cost of reference counting. This is also faster than `epoch::pin()`, perhaps due to better inlining opportunities (?).

The point is, that 1 nanosecond paid on reference counting really doesn't matter at all. If you need maximum performance, it's easy to have it by creating a custom thread-local anyway.

jeehoonkang:
@stjepang thanks for the kind explanation.  Now I prefer the 3rd option.  Let's go for it :)

For safety, I incorrectly assumed that both 1st and 3rd options invoke undefined behavior.  However, the 3rd option seems to raise a specific panic: `HANDLE` is already destroyed.  It's arguably *less* unsafe, I think.

jeehoonkang (approved):
Thanks for a great job!

arthurprs (approved):
Solid proposal. Thanks for putting all this together.

PR #21, opened 2017-11-05
Introduce crossbeam-deque

ghost:
This is a proposal to introduce crate `crossbeam-deque` with a work-stealing Chase-Lev
deque implementation that uses Crossbeam's new epoch-based GC.

If you're using a similar deque implementation right now and have any thoughts, comments, or requests for new functionality, now is a great time to say so. :)

[Rendered](https://github.com/stjepang/rfcs-crossbeam/blob/deque/text/2017-11-05-deque.md)
[Implementation](https://github.com/crossbeam-rs/crossbeam-deque/pull/1)

cc @carllerche @nikomatsakis @cuviper

### DIFF
--- a/text/2017-11-05-deque.md
+++ b/text/2017-11-05-deque.md
@@ -84,8 +84,8 @@ pub enum Steal<T> {
     Empty,
     /// Some data was stolen.
     Data(T),
-    /// Inconsistent state was encountered. Try again.
-    Inconsistent,
+    /// Lost the race for stealing data to another concurrent operation. Try again.
+    Retry,
 }
 ```
 
@@ -112,6 +112,8 @@ None.
 
 # Alternatives
 
+### Hazard pointers
+
 Ultimately, hazard pointers would probably be a better fit for the deque than
 epoch-based garbage collection.
 
@@ -133,6 +135,18 @@ anyone is still using it and continue in the same vein.
 
 At some point in the future, we should definitely experiment with HP-based
 garbage collection.
+
+### Signature of `Stealer::steal`
+
+Instead of returning an explicit enum `Steal<T>`, the `steal` function could also
+return a `Result`, in one of the following two forms:
+
+1. `Result<Option<T>, StealError>`, where `StealError` is equivalent to `Steal::Retry`.
+2. `Result<T, StealError>`, where `StealError` is an enumeration of `Empty` and `Retry`.
+
+Returning a `Result` would allow one to use it with the `?` operator and all
+the other commonly used combinators. However, since stealing is a rather unusual
+and rarely used operation, ergonomics are not of high priority in this situation.
 
 # Unresolved questions
 
--- a/text/2017-11-05-deque.md
+++ b/text/2017-11-05-deque.md
@@ -1,0 +1,150 @@
+# Summary
+
+Introduce crate `crossbeam-deque` with an implementation of the Chase-Lev deque.
+
+# Motivation
+
+Work-stealing deque is a common building block in schedulers, for example in
+[Rayon](https://github.com/nikomatsakis/rayon/) and
+[futures-pool](https://github.com/carllerche/futures-pool).
+Currently, these crates are using the deque from [Coco](https://github.com/stjepang/coco),
+which is an experimental crate that will soon be deprecated in favor of Crossbeam.
+
+`crossbeam-deque` will use the new Crossbeam's epoch-based garbage collector.
+
+# Detailed design
+
+The basis for implementation is the deque from Coco that already had a good run in
+Rayon and consequently in Stylo. However, there will be a few differences:
+
+1. Memory management is done using `crossbeam-epoch`.
+2. The `Worker`/`Stealer` interface is redesigned as `Deque`/`Stealer`.
+3. Several new convenience methods are introduced.
+
+Other than that, the crucial (and the most tricky to implement) methods like
+`push`, `pop`, and `steal` do not depart from the original implementation
+(apart from the changes associated with porting it to Crossbeam's new epoch-based GC).
+
+## The interface
+
+```rust
+/// The "worker" side of the deque.
+pub struct Deque<T>;
+
+unsafe impl<T: Send> Send for Deque<T> {}
+
+impl<T> Deque<T> {
+    /// Creates a new deque.
+    pub fn new() -> Deque<T>;
+
+    /// Creates a new deque with specified minimum capacity.
+    pub fn with_min_capacity(min_cap: usize) -> Deque<T>;
+
+    /// Returns `true` if the deque is empty.
+    pub fn is_empty(&self) -> bool;
+
+    /// Returns the number of elements in the deque.
+    pub fn len(&self) -> usize;
+
+    /// Pushes an element onto the bottom of the deque.
+    pub fn push(&self, value: T);
+
+    /// Pops an element from the bottom of the deque.
+    pub fn pop(&self) -> Option<T>;
+
+    /// Steals an element from the top of the deque.
+    pub fn steal(&self) -> Steal<T>;
+
+    /// Creates a new stealer for the deque.
+    pub fn stealer(&self) -> Stealer<T>;
+}
+
+/// The "stealer" side of the deque.
+pub struct Stealer<T>;
+
+unsafe impl<T: Send> Send for Stealer<T> {}
+unsafe impl<T: Send> Sync for Stealer<T> {}
+
+impl<T> Stealer<T> {
+    /// Returns `true` if the deque is empty.
+    pub fn is_empty(&self) -> bool;
+
+    /// Returns the number of elements in the deque.
+    pub fn len(&self) -> usize;
+
+    /// Steals an element from the top of the deque.
+    pub fn steal(&self) -> Steal<T>;
+}
+
+impl<T> Clone for Stealer<T> { ... }
+
+/// Possible outcomes of a steal operation.
+pub enum Steal<T> {
+    /// The deque was empty.
+    Empty,
+    /// Some data was stolen.
+    Data(T),
+    /// Inconsistent state was encountered. Try again.
+    Inconsistent,
+}
+```
+
+An interesting difference from Coco's API is that a deque is now
+constructed using `Deque::new()` rather than a global function that returns
+a `(Worker<T>, Stealer<T>)` pair.
+Also, there are two ways of creating multiple stealers: you can either create them with
+`Deque::stealer` or clone them - whatever works best for you.
+
+Another addition is the `with_min_capacity` constructor. Deques dynamically grow and
+shrink as elements are inserted and removed. By specifying a large minimum capacity
+it is possible to reduce the frequency of reallocations.
+
+The `steal` method returns a `Steal<T>`. Instead of retrying on a failed CAS, the
+method returns immediately with `Steal::Inconsistent`. This gives schedulers
+fine control over what to do in case of contention.
+
+There is an open [pull request](https://github.com/crossbeam-rs/crossbeam-deque/pull/1)
+with the implementation according to this RFC.
+
+# Drawbacks
+
+None.
+
+# Alternatives
+
+Ultimately, hazard pointers would probably be a better fit for the deque than
+epoch-based garbage collection. In theory, epoch-based GC may leak memory
+indefinitely if a pinned thread is preempted, while hazard pointers give 
+stricter guarantees.
+
+HP-based GC would guarantee that there is a bounded number of
+alive garbage buffers at any time. The main advantage of epoch-based GC is that
+it allows fast traversal through linked data structures, which the Chase-Lev
+deque isn't. This is the fundamental tradeoff between epoch-based and HP-based GC. 
+
+Moreover, HP-based GC would allow us to force perfectly eager destruction of
+buffers: just like `Arc`, but without the overhead of contended reference counting.
+The idea is that when a thread wants to destroy a buffer, it would check
+all hazard pointers to see if it is still used. If there is a thread using it,
+we'd CAS the hazard pointer and set it to null. When that thread notices that
+someone has set its hazard pointer to null, it would take over the responsibility
+of destroying the buffer. Then it would check all hazard pointers to see if
+anyone is still using it and continue in the same vein.
+
+We should definitely experiment with HP-based garbage collection sometime.
+
+# Unresolved questions
+
+What additional methods do we need?
+
+When building `futures-pool`, **[@carllerche](https://github.com/carllerche)**
+wanted a few additional methods:
+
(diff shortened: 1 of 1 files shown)


text/2017-11-05-deque.md
```diff
@@ -0,0 +1,151 @@
...
+constructed using `Deque::new()` rather than using a global function returning
+a `(Worker<T>, Stealer<T>)` pair.
+Also, there are two ways of creating multiple stealers: you can either create each of them
+with `Deque::stealer` or by clone an existing one - whatever works best for you.
+
+Another addition is the `with_min_capacity` constructor. Deques dynamically grow and
+shrink as elements are inserted and removed. By specifying a large minimum capacity
+it is possible to reduce the frequency of reallocations.
+
+The `steal` method returns a `Steal<T>`. Instead of retrying on a failed CAS operation, the
```
jeehoonkang, line 102: I'm slightly inclined to like the following interface better than `Steal`:

```rust
enum StealError {
    LostRace,
}

fn Stealer<T>::steal(&self) -> Result<Option<T>, StealError>;
```

Here, `Ok(Some(val))` means it returns the value `val` and `Ok(None)` means the deque is empty; `Err(LostRace)` means the thread couldn't get any useful information from the deque since it lost a race over the deque.  I think this application of `Result` and `Option` types aligns with the intended use cases of these types.

A drawback is it's a little bit verbose.

What do you think?

jeehoonkang (commented):
I'm almost approving this RFC.  I have only one comment on the function signature of `Stealer::steal()`.

text/2017-11-05-deque.md
ghost: Hmm, I don't know - both versions look fine to me. Let's see what others think.

However, one aspect of your method signature feels a bit unusual. Typically, the empty case is considered to be an error, not a special case of the success case. For example, in the `mpsc` channel, [`try_recv`](https://doc.rust-lang.org/nightly/std/sync/mpsc/struct.Receiver.html#method.try_recv) method has the following signature:

```rust
pub fn try_recv(&self) -> Result<T, TryRecvError>;

pub enum TryRecvError {
    Empty,
    Disconnected,
}
```

So the `Ok` case happens when an element is successfully received, and the `Err` case happens when the channel is either empty or disconnected. In your signature, `steal` results with `Ok` even if the deque is empty.

cuviper:
I don't have strong opinions about the API as far as Rayon is concerned, as long as we can still do the same things and the performance is similar or improved.

The minimum capacity is interesting.  We could expose that indirectly in our `Configuration` builder for the `ThreadPool`, but I'm not sure how we should choose that in general.  Since the `JobRef` is small, only two pointers, we might do well enough with just about a page of memory.

Vtec234 (approved):
Looks great! One thing from me, now that we have a generic `Collector`-based API, maybe it would be good to add a constructor for `deque` from a custom `Collector` or `Handle`?

text/2017-11-05-deque.md
mitchmindtree: > Typically, the empty case is considered to be an error

@stjepang I've always considered this to be a mistake in the case of `try_recv`. The method itself suggests that receiving both `Some` and `None` are expected behaviours and could be considered successful cases - after all, the point of calling `try_recv` instead of `recv` is because we know that there may not be value waiting.

I've found that this tends to make error handling frustrating. I often find myself wanting to write code like this:

```rust
while let Some(msg) = rx.try_recv()? {
    // handle msg
}
```

However the closest I can think of with the current API is significantly more verbose:

```rust
loop {
    match rx.try_recv() {
        Ok(msg) => {
            // handle msg
        },
        Err(mpsc::TryRecvError::Empty) => break,
        Err(err) => return Err(MyError::Disconnected),
    }
}
```

Also, notice how I use a custom error type in the second example too - this is because I rarely find myself wanting to wrap `TryRecvError` in my own error type, as the `Empty` variant is never really an "error" that I want to propagate up through my results in practise as it is almost always expected behaviour.

Anyway, this is just my two cents following personal experience - I could very well be using this wrong! Also I just want to quickly add I really appreciate all the work you're putting into crossbeam at the moment :)

text/2017-11-05-deque.md
```diff
@@ -0,0 +1,151 @@
...
+At some point in the future, we should definitely experiment with HP-based
+garbage collection.
+
+# Unresolved questions
+
+When building `futures-pool`, **[@carllerche](https://github.com/carllerche)**
+wanted a few additional methods:
+
+1. [A method that steals more than one value at a time.](https://github.com/stjepang/coco/issues/11#issuecomment-339785208)
+2. [A `steal_when_greater` method.](https://github.com/stjepang/coco/issues/10#issuecomment-339785563)
```
carllerche, line 157: This is pretty low in priority IMO. As you pointed you can approximate this with the `len` API.

text/2017-11-05-deque.md
```diff
@@ -0,0 +1,151 @@
...
+
+At some point in the future, we should definitely experiment with HP-based
+garbage collection.
+
+# Unresolved questions
+
+When building `futures-pool`, **[@carllerche](https://github.com/carllerche)**
+wanted a few additional methods:
+
+1. [A method that steals more than one value at a time.](https://github.com/stjepang/coco/issues/11#issuecomment-339785208)
```
carllerche, line 156: Definitely interesting, but I agree that this isn't a requirement for moving forward and could be explored later.

carllerche (approved):

text/2017-11-05-deque.md
ghost: @mitchmindtree Thanks for chiming in - you're presenting a very good argument!

While I agree that returning `Ok(None)` instead of `Err(Empty)` would make implementing some patterns easier, it's still slightly odd to return an `Ok` when a receive operation fails.

I mean, `try_recv` didn't *recv* a message, but it still returned an `Ok`? Hmmm... :)

Also, what if you want to receive a message from a channel, while asserting that it's not disconnected nor empty? You'd have to write:

```rust
let msg = rx.try_recv().unwrap().unwrap();
```

Aren't there too many layers of wrapping?

Perhaps this is a sign that we should keep the `Steal` enum instead of returning a `Result`? The `steal` operation is pretty unusual, so enumerating the entire list of cases on each call would make the intent of the code more obvious.

Furthermore, concurrent deques and, consequently, calls to `steal` are not very common in Rust programs (unlike channels and `try_recv`), so I believe there isn't a strong incentive to make the method as ergonomic as possible.

mitchmindtree: > I mean, try_recv didn't recv a message, but it still returned an Ok? Hmmm... :)

Hmmm I find it interesting that you bring this up as a point against, as this seems like the intuitive behaviour to me 😄 

In my view, `try_recv` exists in order to avoid blocking when no values are present, so receiving `None` is often the expected, "successful" (`Ok`) result - I haven't personally come across a case yet where I'd describe the channel being empty as an "error" in the flow of my code or unexpected behaviour, as it seems to me this is the point of using `try_recv` instead of `recv` in the first place.

> `let msg = rx.try_recv().unwrap().unwrap();`
>
> Aren't there too many layers of wrapping?

In this case I'd argue that perhaps the user should be using `recv` if they are expecting a value to exist and `panic!`ing otherwise, e.g. `rx.recv().unwrap()`?

On the other hand, if a user really did come across a use-case for this, I'd imagine this would be more common in practise `rx.try_recv()?.unwrap()`.

> Perhaps this is a sign that we should keep the Steal enum instead of returning a Result?

Admittedly I'm not very familiar with this proposal - I just happened to come across your reference to the `try_recv` API and it sparked some frustrating memories :) Perhaps I'd be better off raising this as an alternative ergonomic approach in #22?

jeehoonkang:
I agree with @Vtec234 on the interface: I'd like to have a `Collector`-based API for crossbeam crates.  @stjepang Would you please say some words on the interface in the unresolved questions section?

Except for that and the interface of `steal()`, I'm basically approving this RFC.  Thanks!

ghost:
@jeehoonkang 

I still believe we should keep the `Steal<T>` enum. Here's why.

When you call `try_recv()` to send an element into a channel, there are three outcomes:

1. An element was received.
2. The channel was empty.
3. The channel is disconnected.

The first case is clearly a success (`Ok`), and the third one is clearly a failure (`Err`). We'll disagree on the second case, though. :) I see it as a failure because the receive operation failed to receive an element. You might say it's a success because the channel wasn't disconnected. That's also a valid position.

The situation is even more confusing with concurrent deques. When you call `steal()`, there are three possible outcomes:

1. An element was stolen.
2. The deque was empty.
3. Inconsistent state observed.

Again, the first case is a clear success. The second case is, again, a success or a failure depending on how you think about it. Regarding the third case, you might say it's a failure because essentially nothing happened. But, I would argue, the third case can also be viewed as a "neutral" outcome (nothing happened, you should just retry), while the second case is a failure (failed to steal an element because the deque is empty).

In the end, whatever we decide is a success (`Ok`) and a failure (`Err`), someone will find the choice unintuitive. And I think this impairs readability, so when you see a line like this one, it's not completely obvious what exactly happens:

```rust
let msg = stealer.steal()?;
```

But if you're matching on an enum, there's no confusion at all:

```rust
let msg = match stealer.steal() {
    Steal::Empty => None,
    Steal::Data(d) => Some(d),
    Steal::Inconsistent => return Err(whatever),
};
```

Of course, this is much more verbose.

I appreciate @mitchmindtree's concerns about ergonomics. Since channels are very [popular](https://github.com/search?l=Rust&q=try_recv&type=Code&utf8=%E2%9C%93), it makes sense to discuss how to make their use as ergonomic as possible.

But concurrent deques are different - they are used very [infrequently](https://github.com/search?utf8=%E2%9C%93&q=%22steal%3A%3Aabort%22+in%3Afile+language%3ARust+extension%3A.rs&type=Code), so in their case, I believe we should pursue clarity over ergonomics instead.

There is already some precedent for choosing an explicit enum instead of relying on `Result`: an [ancient](https://github.com/alexcrichton/rust/blob/a70f9d7324a91058d31c1301c4351932880d57e8/src/libstd/rt/deque.rs#L98) version of deque in Rust's runtime, the [deque](https://github.com/kinghajj/deque/blob/master/src/lib.rs#L89) crate, current [Crossbeam](https://docs.rs/crossbeam/0.2.10/src/crossbeam/sync/chase_lev.rs.html#95), and the mpsc [implementation](https://github.com/rust-lang/rust/blob/bd7ce71aabfcd851b94a097060643c056a9b0659/src/libstd/sync/mpsc/mpsc_queue.rs#L33) (see [here](https://github.com/rust-lang/rust/blob/bd7ce71aabfcd851b94a097060643c056a9b0659/src/libstd/sync/mpsc/shared.rs#L196)) in the standard library. I think it worked out quite well.

jeehoonkang:
> In the end, whatever we decide is a success (`Ok`) and a failure (`Err`), someone will find the choice unintuitive.

I thought the empty case is definitely a success, but it turns out that someone (you, at least) disagrees with me.  In that case, I agree with you in choosing clarity over ergonomics or doing someone's "right" thing.  Let's leave the `Stealer::steal()`'s interface as-is.

Anyway, I'd like to ask if you could put @mitchmindtree's and my argument for the changes to the interface in the RFC.  Here are two more cents from me.  I think `Result<Option<T>, StealError>` more clearly reveals the relation between `Deque::pop()/steal()` and `Stealer::steal()`: the result type of the latter is similar to that of the former, but it may have lost in a race (`StealError`).  That relation makes me to think that the empty case is actually not an error for both `Deque::pop()/steal()` and `Stealer::steal()`.

glaebhoerl:
This seems quite similar to the questions around `Future`s and the `?` operator (I think this was discussed in the RFC for the `Try` trait?), which similarly had a 3-way split when `await`ing: got a value, got an error, or got a "try again next time", with the debate likewise being around whether that last one should count as an error, or what.

(Going by memory; I should dig it up but I'm lazy.)

Anyway, I think we could use a similar solution here as what I remember the resolution there being: go with the 3-way enum, *and* also have convenience methods on it to translate it to the different possible `Result` types, based on which things are to be considered errors in the given situation. This retains most of the convenience of using a `Result`, and arguably helps readability as well. So you'd write something like (specific names only for the sake of demonstration!) `stealer.steal().consistent()?` (resulting in an `Option<T>`) where only inconsistency is an error, or `stealer.steal().stole_something()?` (resulting in a `T`) where emptiness is as well.

ghost:
> Anyway, I'd like to ask if you could put @mitchmindtree's and my argument for the changes to the interface in the RFC.

Will add to the text.

> Anyway, I think we could use a similar solution here as what I remember the resolution there being: go with the 3-way enum, and also have convenience methods on it to translate it to the different possible Result types, based on which things are to be considered errors in the given situation.

Sounds like a great idea - best of both worlds!

I've managed to dig up a few relevant links to similar discussion around `Poll` in `futures`: [[1]](https://docs.rs/futures/0.1.17/futures/future/trait.Future.html#return-value) [[2]](https://github.com/alexcrichton/futures-rs/issues/141) [[3]](https://github.com/alexcrichton/futures-rs/issues/141) [[4]](https://github.com/rust-lang/rfcs/blob/master/text/1859-try-trait.md#using--with-types-other-than-result) [[5]](https://github.com/rust-lang/rfcs/pull/1859#issuecomment-279187967) [[6]](https://github.com/rust-lang/rfcs/pull/1859#issuecomment-295569693).

@alexcrichton Do you perhaps have an opinion here? Here's a quick explanation of the dilemma. Currently, we have the following API for the steal operation in a concurrent deque:

```rust
impl<T> Stealer<T> {
    /// Steals an element from the top of the deque.
    pub fn steal(&self) -> Steal<T>;
}

/// Possible outcomes of a steal operation.
pub enum Steal<T> {
    /// The deque was empty.
    Empty,
    /// Some data was stolen.
    Data(T),
    /// Inconsistent state was encountered. Try again.
    Inconsistent,
}`
```

We're wondering whether method `steal` should return `Result<Option<T>, Inconsistent>` instead. This is kinda similar to `Poll` in `futures`. Although in our case ergonomics around error handling are a low-priority concern, I'm still wondering what is *the right* thing to do here. What do you think?

alexcrichton:
@stjepang the primary use case for `Result` in `poll` for futures was to enable `?` syntax, and it sounds like that may not be a primary concern here perhaps? (no need to enable the ? syntax that is).

We actually started out with `enum Poll<T> { ... }` so I'd probably go for the bare `enum` route myself if you don't need to use `?` that much

ghost:
Let's go with the explicit enum then. I've added a remark on `Steal` vs `Result` to the *Alternatives* section.

Also, after chatting with @jeehoonkang, I've renamed `Inconsistent` to `Retry`.

jeehoonkang (approved):

PR #22, opened 2017-11-09
Introduce crossbeam-channel

ghost:
Introduce crate `crossbeam-channel`, which aims to be an upgrade over `std::sync::mpsc` from the standard library in pretty much all aspects: features, convenience, and performance.

[Rendered RFC](https://github.com/stjepang/rfcs-crossbeam/blob/channel/text/2017-11-09-channel.md)
[Prototype implementation](https://github.com/stjepang/channel)
[Selection macro](https://github.com/stjepang/channel/pull/3)
[Benchmarks](https://github.com/stjepang/channel/tree/master/benchmarks)

Graphs

cc @jdm @SimonSapin @asajeffrey - we've chatted about this in `#servo`.
cc @alexcrichton @aturon @mgattozzi - talked about this at RustConf.
cc @BurntSushi - this might be a replacement for `chan`.
cc @insanitybit - finally! :)

### DIFF
--- a/text/2017-11-09-channel.md
+++ b/text/2017-11-09-channel.md
@@ -1,0 +1,936 @@
+# Summary
+
+Introduce crate `crossbeam-channel`, which aims to be an upgrade over
+`std::sync::mpsc` from the standard library in pretty much all aspects:
+features, convenience, and performance.
+
+The prototype implementation can be found [here](https://github.com/stjepang/channel).
+Also, check out [benchmarks](https://github.com/stjepang/channel/tree/master/benchmarks)
+to see how it fares against `std::sync::mpsc` and other crates offering
+different kinds of channels and queues.
+
+# Motivation
+
+The design of the new channel is driven by an attempt to answer three questions:
+
+1. How to fix the shortcomings of `std::sync::mpsc`?
+2. How to achieve the flexibility of Go's channels, and especially its *select*?
+3. How to consolidate the variety of Java's concurrent queues into a unified channel interface?
+
+## Shortcomings of `std::sync::mpsc`
+
+I've compiled a list of some of the problems our fellow Rust users have been complaining about:
+
+1. StackOverflow user [asking](https://stackoverflow.com/questions/40384274/rust-mpscsender-cannot-be-shared-between-threads)
+   why [`Sender`](https://doc.rust-lang.org/std/sync/mpsc/struct.Sender.html)
+   doesn't implement `Sync`.
+2. In Rust-related channels on irc.mozilla.org, people regularly
+   come in asking for [MPMC](https://botbot.me/mozilla/rust/search/?q=mpmc) or
+   [SPMC](https://botbot.me/mozilla/rust/search/?q=spmc) channels and queues.
+3. A Reddit user is [asking](https://www.reddit.com/r/rust/comments/6remch/thoughts_on_avoiding_busy_waiting_for_threads/)
+   for help with multithreading. Complains about "pretty confusing" API and "apparently mediocre"
+   implementation of [`Select`](https://doc.rust-lang.org/nightly/std/sync/mpsc/struct.Select.html).
+   There is also some talk about MPMC channels.
+4. Another Reddit user [wants](https://www.reddit.com/r/rust/comments/3mpke2/does_rust_has_something_like_channels_in_go/cvhs5kk/)
+   SPMC channels. The original post is asking about Go-like channels in Rust.
+5. An annoyance/wart [reported](https://github.com/rust-lang/rust/issues/12902)
+   on Rust's issue tracker: The `select!` macro doesn't work if you try to reference
+   a `Receiver` in a struct.
+6. `Sender` not being `Sync` is a common stumbling block for users of Hyper, as
+   [explained](https://github.com/rust-lang/rfcs/pull/1299#issuecomment-146361678)
+   by [**@seanmonstar**](https://github.com/seanmonstar).
+7. [**@jonhoo**](https://github.com/jonhoo)
+   [points out](https://github.com/rust-lang/rust/issues/27800#issuecomment-313455851)
+   that channel selection in Rust is not fair like in Go. The list of cases
+   in a `select!` should be randomly permuted each time.
+8. [**@alexcrichton**](https://github.com/alexcrichton)
+   [says](https://github.com/rust-lang/rust/issues/12902#issuecomment-319510050)
+   the `select!` macro is defacto deprecated nowadays. It's probably alive only
+   because Servo is still using it.
+9. Judging by upvotes given to a [comment](https://github.com/rust-lang/rust/issues/27800#issuecomment-140966321)
+   on GitHub, people really really want Go-like channels with easy selection.
+10. A [plea](https://github.com/rust-lang/rust/issues/27800#issuecomment-270583295)
+    for the ability to select across dynamic lists of channels.
+11. The selection interface [doesn't support](https://github.com/rust-lang/rust/issues/27800#issuecomment-300880631)
+    specifying a timeout.
+12. A Rocket user [wants](https://www.reddit.com/r/rust/comments/6t8sdl/help_communicating_bidirectionally_with_rocket_in/)
+    to pass a `Receiver` to `.manage(...)`, but this doesn't work because `Receiver` is not `Sync`.
+13. [**@alexcrichton**](https://github.com/alexcrichton)
+    [lays out](https://github.com/rust-lang/rust/pull/42397#issuecomment-315867774)
+    a list of regrets about the current design of `std::sync::mpsc`.
+14. Since `Receiver` is not `Sync`, people [work around](https://github.com/pingcap/tikv/pull/637/files)
+    the problem by wrapping it in a `Mutex`; for example: `Arc<Mutex<Receiver<Option<T>>>>`.
+15. Yet another [example](https://github.com/servo/servo/blob/38f4ae80c4b456b89ee33543c8c6699501696c9c/components/script/dom/paintworkletglobalscope.rs#L234)
+    in Servo: wrapping `WorkletExecutor` in a `Mutex` because `Sender` is not `Sync`.
+16. [**@jdm**](https://github.com/jdm)
+    [says](https://mozilla.logbot.info/servo/20170719#c716258)
+    having a `.len()` method on channels would be useful for profiling Servo.
+17. In an effort to get rid of unstable features, Servo is
+    [looking for](https://github.com/servo/servo/issues/5286)
+    an alternative to `mpsc_select`.
+18. [`ipc-channel`](https://github.com/servo/ipc-channel)
+    depends on `mpsc_select`, so it
+    [cannot be used](https://github.com/servo/ipc-channel/issues/118)
+    with stable Rust.
+19. [Rust severely disappoints](http://esr.ibiblio.org/?p=7294) ESR:
+    *Not only that, but it seems the CSP implementation in Rust – the
+    most tractable concurrency primitive – has limits that make it unusable
+    for NTPsec’s purposes (only selection over a static set of channels is possible)
+    and there is some danger that it might be removed entirely!*
+
+I also wrote a lengthy critique on `std::sync::mpsc` in a
+[blog post](https://stjepang.github.io/2017/08/13/designing-a-channel.html),
+with suggestions for how to design a new, better channel.
+
+Here's a summary of all the shortcomings of `std::sync::mpsc`:
+
+1. Channels are not MPMC (they don't support multiple `Receiver`s).
+2. The bounded variant (`sync_channel`) is slow and scales poorly due to internal lock contention.
+3. `Sender`s and `Receiver`s are not `Sync`.
+4. There is no `send_timeout` method (useful when sending data into a full bounded channel).
+5. [`Sender`](https://doc.rust-lang.org/std/sync/mpsc/struct.Sender.html)
+   and [`SyncSender`](https://doc.rust-lang.org/std/sync/mpsc/struct.SyncSender.html)
+   are distinct types, which duplicates the interface somewhat.
+6. `select!` can only select over *receive* operations, but not over *send* operations.
+7. `select!` does not have a clear path towards stabilization.
+8. Dynamic selection (using [`Select`](https://doc.rust-lang.org/nightly/std/sync/mpsc/struct.Select.html))
+   is unsafe and unergonomic, as is evident by
+   [code](https://github.com/servo/servo/blob/fa319170ebb34afcdfc120b7c3e47fe5b1c21210/components/script/script_thread.rs#L930)
+   in Servo.
+9. Channel selection is not fair (select cases are always fired in the same order).
+10. The `select!` interface comes with some surprising warts.
+11. It would be nice to have `len` method on `Sender`s and `Receiver`s.
+
+So far, the most popular brand of band-aid for patching some of these problems has been
+the [chan](https://github.com/BurntSushi/chan) crate, which comes with MPMC channels.
+Unfortunately, it also brings some problems of its own, poor performance being the
+most pronounced one.
+
+## A look at Go's channels
+
+Channels and selection are the core building blocks of Go programs, and a huge
+reason for the success of language.
+
+Go's channels come in two flavors: unbuffered (zero capacity) and buffered (positive,
+fixed capacity) channels. There are no unbounded channels built in.
+
+Sometimes they get [criticized](http://www.jtolds.com/writing/2016/03/go-channels-are-bad-and-you-should-feel-bad/)
+for being slow, which is probably a bit too harsh. If we
+[compare](https://github.com/stjepang/channel/tree/master/benchmarks)
+them to Rust's `std::sync::mpsc`, they actually perform quite well.
+
+Where Go really shines, however, is the versatility of *select*:
+
+```go
+// Select over two operations on a zero-capacity channel named `match`.
+select {
+
+// Receiving message `peer` from channel `match`.
+case peer := <-match:
+    fmt.Printf("%s received a message from %s.\n", name, peer)
+
+// Sending message `name` into channel `match`.
+case match <- name:
+    fmt.Printf("Someone has matched with %s.\n", name)
+}
+```
+
+Here we are declaring a receive and a send operation, and blocking until
+exactly one of them succeeds. This is something `std::sync::mpsc` simply
+cannot do. In fact, it's difficult to find a language (or a library)
+that does offer something equivalent.
+
+On the other hand, dynamic selection (declaring select cases dynamically
+rather than statically) can be [a bit tricky](https://stackoverflow.com/questions/19992334/how-to-listen-to-n-channels-dynamic-select-statement)
+to do (and probably slow), but Go's *select* is already a high bar to meet.
+
+There has been some effort by [Dmitry Vyukov](http://www.1024cores.net/) to improve the performance of Go's
+channels by avoiding contention on mutexes (often referred to as *lock-free* channels,
+although they aren't exactly, [strictly speaking](https://en.wikipedia.org/wiki/Non-blocking_algorithm#Lock-freedom)):
+
+* (2013/08 - present) [Implementation](https://codereview.appspot.com/12544043/) submitted for review.
+* (2014/02) Initial [proposal](https://groups.google.com/forum/#!msg/golang-dev/5zcEng3yvaU/wfosgbaKuEgJ) for Go 1.3.
+* (2014/02) The design [document](https://docs.google.com/document/d/1yIAYmbvL3JxOKOjuCyon7JhW4cSv1wy5hC0ApeGMV9s/pub).
+* (2014/07) An [update](https://groups.google.com/forum/#!searchin/golang-dev/lock-free|sort:relevance/golang-dev/0IElw_BbTrk/FtV82Tr3S00J).
+* (2014/10) The [issue](https://github.com/golang/go/issues/8899) on GitHub.
+* (2016/03) [Discussed](https://groups.google.com/forum/#!searchin/golang-nuts/lock-free|sort:relevance/golang-nuts/LM648yrPpck/o1rR6AUhAwAJ)
+  as a possible answer to the aforementioned [critique](http://www.jtolds.com/writing/2016/03/go-channels-are-bad-and-you-should-feel-bad/).
+* (2017/04) Another [update](https://github.com/golang/go/issues/8899#issuecomment-292491545).
+* (2017/05) [Brought up](https://github.com/golang/go/issues/20351#issuecomment-301895677) again as a possible solution to performance regressions in Go 1.8.
+
+This new implementation has been in a limbo state for several years already. It never
+fully materialized, but it was also neither abandoned nor given a red light.
+It might still end up in Go's mainline at some point in the future. We'll see.
+
+## A look at Java's concurrent queues
+
+Java has a quite extensive list of data structures in [`java.util.concurrent`](https://docs.oracle.com/javase/8/docs/api/index.html?java/util/concurrent/package-summary.html).
+We're aspiring to have many of those implemented in Rust as part of the Crossbeam project.
+
+Java's concurrent queues:
+
+1. [`ArrayBlockingQueue`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ArrayBlockingQueue.html)
+   is a bounded blocking queue backed by an array. This is similar to `std::sync::mpsc::sync_channel`.
+2. [`ConcurrentLinkedQueue`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentLinkedQueue.html)
+   is an unbounded queue backed by a linked list. This is similar to `std::sync::mpsc::channel`.
+3. [`DelayQueue`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/DelayQueue.html)
+   is an unbounded blocking queue of delayed elements. An element can only be taken after
+   its delay has expired. This is a special kind of priority queue.
+4. [`LinkedBlockingQueue`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/LinkedBlockingQueue.html)
+   is an optionally-bounded blocking queue based on a linked list. In bounded form, it's
+   rarely useful because `ArrayBlockingQueue` is probably a better choice anyway.
+5. [`LinkedTransferQueue`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/LinkedTransferQueue.html)
+   is similar to `ConcurrentLinkedQueue`, except it also has the ability to block on
+   consumers.
+6. [`PriorityBlockingQueue`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/PriorityBlockingQueue.html)
+   is an unbounded blocking priority queue. This is just a thin layer on top of a
+   typical priority queue.
+7. [`SynchronousQueue`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/SynchronousQueue.html)
+   is a zero-capacity (also known as *rendezvous*) queue. This is equivalent to
+   `std::sync::mpmsc::sync_channel(0)`.
+
+Rust's `std::sync::mpsc` already encompasses the feature sets of (1), (2), (5), and (7),
+except for the ability to have multiple consumers and things that don't fit Rust's
+ownership model. For example, all Java's concurrent queues have the `peek()` method,
+which returns a reference to an element but doesn't remove it from the queue. This
+is obviously a red flag in the Rust world.
+
+Queues (3) and (6) are a bit special and depart from the domain of typical channels.
+Queue (4) is not difficult to implement, but at the same time, it also doesn't seem
+particularly useful. Therefore, those three queues are not the focus of `crossbeam-channel`.
+
+# Detailed design
+
+## Design goals
+
+All shortcomings of `std::sync::mpsc` enumerated in the *Motivation* section must be fixed.
+`crossbeam-channel` strives to look simple and stupid on the outside,
+but be ambitiously powerful inside. The channel will:
+
+1. Be an MPMC channel (allow multiple `Sender`s and multiple `Receiver`s).
+2. Avoid locks and be much faster than `std::sync::mpsc` in the bounded case (the `sync_channel` case).
+3. Not be significantly slower than `std::sync::mpsc` in the unbounded case.
+4. Have the full spectrum of `try_send`/`send`/`send_timeout`/`try_recv`/`recv`/`recv_timeout` methods.
+5. Have convenience methods like `capacity`, `len`, and `is_empty`.
+6. Have `Senders` and `Receivers` that implement `Sync`.
+7. Have a simple interface without special types like `SyncSender`.
+8. Support selection over both send and receive operations.
+9. Allow selection with a timeout.
+10. Have easy and safe dynamic selection.
+11. Have selection without the warts of `select!`.
+12. Have fair(er) selection.
+
+## The interface
+
+Constructors:
+
+```rust
+pub fn unbounded<T>() -> (Sender<T>, Receiver<T>);
+pub fn bounded<T>(cap: usize) -> (Sender<T>, Receiver<T>);
+```
+
+Sender:
+
+```rust
+pub struct Sender<T> { ... }
+
+unsafe impl<T: Send> Send for Sender<T> {}
+unsafe impl<T: Send> Sync for Sender<T> {}
+
+impl<T> Sender<T> {
+    pub fn try_send(&self, value: T) -> Result<(), TrySendError<T>>;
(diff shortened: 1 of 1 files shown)


text/2017-11-09-channel.md
```diff
@@ -0,0 +1,944 @@
...
+pub struct IntoIter<T> { ... }
+impl<T> Iterator for IntoIter<T> { ... }
+```
+
+Selection:
+
+```rust
+pub mod select {
+    pub fn send<T>(tx: &Sender<T>, value: T) -> Result<(), T>;
+    pub fn recv<T>(rx: &Receiver<T>) -> Result<T, ()>;
```
cramertj: Super minor nit, but I'd use a ZST error type like `struct NotReady;` or something here, rather than `()`.

ghost:
In the Reddit [comment thread](https://www.reddit.com/r/rust/comments/7bszr4/crossbeam_rfcs_introduce_crossbeamchannel/dpl22cw/), [/u/lurebat](https://www.reddit.com/user/lurebat) argues that we should have explicit `Select` structs instead of hidden thread-local magic.

> I just don't like state I can't see. Usually when I call a free function with an argument the only thing I expect to modify is that argument. But here the function doesn't even take a mutable reference and it modified a hidden state that I can't even access. I feel that it's more rustic to go with the explicit option by default, and I don't see why not have it here.

I'm sympathetic to the argument.

Here's an example of selection using hidden thread-local state:

```rust
loop {
    if let Ok(msg) = select::recv(&rx1) {
        println!("{}", msg);
        break;
    }
    if let Ok(msg) = select::recv(&rx2) {
        println!("{}", msg);
        break;
    }
    if select::timeout(Duration::from_millis(100)) {
        break;
    }
}
```

And this is what it would like with an explicit `Select` struct:

```rust
let mut sel = Select::with_timeout(Duration::from_millis(100));
loop {
    if let Ok(msg) = sel.recv(&rx1) {
        println!("{}", msg);
        break;
    }
    if let Ok(msg) = sel.recv(&rx2) {
        println!("{}", msg);
        break;
    }
    if sel.timed_out() {
        break;
    }
}
```

We could require the `sel` variable to be `mut` or not - it doesn't really matter because we can use `Cell`s anyway. But perhaps requiring `mut` makes it clearer that there is some kind of mutable state?

Note that this way we can also make timeouts much less magical.

Any thoughts? Please upvote if you think we should go with explicit `Select` struct, and downvote if you prefer hidden thread-local state.

nikomatsakis:
> We could require the sel variable to be mut or not - it doesn't really matter because we can use Cells anyway. But perhaps requiring mut makes it clearer that there is some kind of mutable state?

One advantage to not requiring mutable is that people may want to put the `Select` struct into some sort of context that gets copied around. If you require `mut`, then they would need to use a `RefCell` or some such thing somewhat needlessly.

aturon:
I wish I had time to give this a proper review, but meantime, just wanted to say: thanks so much, @stjepang (and team), for what you've done for Crossbeam! I couldn't be happier.

bstrie:
@stjepang I'm in favor of the explicit struct approach, though it seems as though we could also offer the implicit approach as well if people are worried about verbosity, yes?

As for the need to explicitly `break`, is there some possible simplification to be had in association with the `catch` RFC? Keep in mind that I am **completely** unclear on the current state of that RFC (both in design and implementation), but would you find something like the following a bit nicer?
```rust
catch {
    loop {
        sel.recv(&rx1, |msg|
            println!("{}", msg);
        )?;
        sel.timed_out()?;
    }
}
```
If nothing else, this would seem to have the benefit of warning the user if a `?` is forgotten, via must_use.

TimNN:
I'm personally in favor of the explicit struct approach as well, however I believe that a macro, which also enforces the loop structure, could work quite well and wouldn't be too complicated, for example:

https://gist.github.com/TimNN/669d030cc99e4dcefc2b5bc6203a7019

droundy:
I notice that the select algorithm as described (in the "Rendered" link) is not actually fair.  Is there a strong performance reason not to go with a fair implementation? I would expect a fair implementation to either check just one option each time through the loop (instead of about half) or to check all the options in a random order at the beginning of the loop.

The unfairness would come around if you have a select loop with many options, and a pair of them are always triggered simultaneously.  If the two that are triggered simultaneously are next to each other in the loop, then the first one will be chosen far more often than the second one.  Maybe this is unimportant, and maybe the performance cost of making it fair is not worthwhile.  But it seems worth investigating.

ebarnard:
To make the explicit struct approach a bit neater and remove the need for explicit `break`s something like this might work.

If you have a mutable reference or are constructing on the spot:
```rust
for sel in Select::with_timeout(Duration::from_millis(100)) {
    if let Ok(msg) = sel.recv(&rx1) {
        println!("{}", msg);
    }
    if let Ok(msg) = sel.recv(&rx2) {
        println!("{}", msg);
    }
}
```
or if you needed to store a `Select` in a struct behind an & reference:
```rust
let select = Select::with_timeout(Duration::from_millis(100));

for sel in select.iter() {
    if let Ok(msg) = sel.recv(&rx1) {
        println!("{}", msg);
    }
    if let Ok(msg) = sel.recv(&rx2) {
        println!("{}", msg);
    }
}
```

droundy:
@ebarnard I like that idea.  The fewer invariants the API forces the user to manage (in this case, break on success of any rule), the better!

TimNN:
The `for` API looks nice, however has the disadvantage that you cannot return a value using `break $something` (unless Rust has changed recently and I didn‘t notice).

gmorenz:
Could we allow for both the `for` api, and the `loop/break` with struct API?

Alternatively you should still be able to wrap it in a function and use return to emulate `break $something`.

ebarnard:
@TimNN @gmorenz 

You can return a value from a loop using a similar construction to before.

```rust
let mut select = Select::new()

let msg = 'select: loop {
    // Panics if you forget to break
    let sel = select.next().unwrap()

    for rx in &receivers {
        if let Ok(msg) = sel.recv(rx) {
            break 'select msg;
        }
    }
};
```

ghost:
@bstrie 
> I'm in favor of the explicit struct approach, though it seems as though we could also offer the implicit approach as well if people are worried about verbosity, yes?

Well, I'd rather choose one or the other. Having two separate, but marginally different interfaces for doing the same thing is kinda silly, isn't it? :) Let's just go with the explicit struct approach.

> As for the need to explicitly `break`, is there some possible simplification to be had in association with the `catch` RFC? Keep in mind that I am **completely** unclear on the current state of that RFC (both in design and implementation), but would you find something like the following a bit nicer?

Interesting idea! While it does look neat, I'm worried that it just sort of pushes the problem from one place to another - now we don't have explicit `break`s, but still have to write `catch` and `?` instead. Also, the control-flow graph now loses important information. See below.

@ebarnard 
> To make the explicit struct approach a bit neater and remove the need for explicit breaks something like this might work.

Very nice, but this way the control-flow graph loses information - it doesn't know that the body of a successful case is executed just once anymore.

Consider the following:

```rust
let output;
let mut sel = Select::new();
loop {
    if let Ok(msg) = sel.recv(&rx1) {
        output = msg.to_string();
        break;
    }
    if let Ok(msg) = sel.recv(&rx2) {
        output = msg.to_string();
        break;
    }
}
println!("{}", output);
```

This code compiles because Rust knows that `output` will be initialized exactly once. From the control-flow-graph, it can deduce that the body of each of the `if`s will be executed at most once, and that exactly one `if` will succeed exactly once.

If you rewrite the same example using `for sel in select.iter()`, it will not compile.

ebarnard:
@stjepang 
That's a shame, I hadn't thought about that. It's still possible to use the `for` based api with a loop if you want to return a value (see above).

However, in the case you only want to perform an operation on each received value, or there's a timeout, or some other fallible assignment the for loop approach is much nicer to use.

I suppose there's a cost to explaining both ways of using the select api though.

ghost:
@nikomatsakis 
> One advantage to not requiring mutable is that people may want to put the `Select` struct into some sort of context that gets copied around. If you require `mut`, then they would need to use a RefCell or some such thing somewhat needlessly.

Good point! Although, I wonder why would someone attempt to store a `Select` anywhere else other than one line above the `loop { .. }`. Seems like an unusual and potentially dangerous thing to do.

Also, I'm feeling a bit uneasy about allowing to have multiple references to the same `Select` in two completely different places and then using them.

@TimNN 

> I believe that a macro, which also enforces the loop structure, could work quite well and wouldn't be too complicated, for example:

Looks pretty good! A question: would it be possible to allow optional commas between select cases? Putting an unexpected comma (or forgetting one) in macros can blow up with poor compilation error messages.

Other than that one, I don't see any quirks in the macro. Are there any I'm missing? :)

Also, we should probably also support the following syntax: `send(tx, mut s)` so that the returned value in `Err(v)` is reacquired (`s = v`).

rpjohnst:
> Instead of just one universal `Sender`/`Receiver` pair, there could be more types for different, better-optimized kinds of channels. For example, we could have `unbounded::Sender`/`unbounded::Receiver`, or `spsc::Sender`/`spsc::Receiver`, or maybe even `oneshot::Sender`/`oneshot::Receiver`.

I like this alternative quite a bit. It eliminates the redundant monomorphizations, it eliminates the runtime dispatch on flavor, and it allows for non-`Sync` variants that not only achieve the same performance as `std::sync::mpsc`, but enforce it at compile time.

It also doesn't preclude a dynamic wrapper that behaves the way the current prototype does, though that would potentially make things more confusing as it would provide multiple types for the same purpose.

droundy:
@rpjohnst The problem I see with having many different channel versions is that it vastly impedes the use of channels in library interfaces.  A library author either needs to create N different versions of the API, or needs to decide which variant is required by their users.  I expect the cost of the dynamic dispatch is negligible (testing would be wise), and you get a simple API that allows the users of libraries to patch them together however they like.

rpjohnst:
It would of course also be possible to solve that problem by making `Sender` and `Receiver` traits, and letting libraries work with those.

johncf:
I don't understand why select has to have such complicated API. I have a much simpler proposal (which might be infeasible since I have no knowledge of the internals):

```rust
let select = rx1.join(&rx2).join(&rx3).with_timeout(8);
select.wait();
if let Some(msg) = rx1.recv_nb() { ... }
else if let Some(msg) = rx2.recv_nb() { ... }
...
else {
    // timed out
}
```
(Note: `_nb` stands for non-blocking)

Edit: clarify

jdm:
@johncf That interface does not provide any fairness guarantees, so you could end up starving the second receiver if there's a steady stream on the first one and you're using the select in a loop.

johncf:
@jdm Ah, then why not use `select.recv(&rx1)` as proposed before instead of `rx1.recv_nb()` ?

bstrie:
@TimNN's macro looks lovely, and seems far less fraught than the analogous macro from `chan`. Does anyone see anything immediately wrong with it? Perhaps it will end up less nice if we have to account for unforeseen edge cases. :P 

@stjepang , note that when I say that I prefer the explicit `Select` struct, what I *really* mean is that I just don't like global state/TLS. :) I, and I assume many others, would likely be totally fine with the select machinery being "behind-the-scenes" via a macro, especially if you're concerned about the risk of further misuse of the API. (After all, all the machinery used by a macro would be public, and so people could still choose to use the explicit `Select` struct if they so desired, they just wouldn't *have* to.)

TimNN:
@stjepang: I don't think there is an easy way for *optional* commas, but allowing an arbitrary number of commas (`$(,)*`) is easily possible (I've updated the gist).

TimNN:
@stjepang 

> Also, we should probably also support the following syntax: send(tx, mut s) so that the returned value in Err(v) is reacquired (s = v).

The `send(tx, $ident)` form of the current macro already recovers the value of `$ident` in case of an error (although that currently requires `$ident` to be `mut`, but that should be easy to fix: The value of `$ident` moves into the loop anyway and is unusable afterwards, so adding an `let mut $ident = $ident` before the loop should be easy enough)

johncf:
Here is a refinement of the idea I proposed before, now having APIs for sending and possibly recovering failed-to-send messages.

https://gist.github.com/johncf/81778f23da91ae928e0ee7e92fda1b48

KasMA1990:
Reading through this RFC, I was quite confused by what exactly `select` is supposed to do. Granted, using channels is new to me, but it wasn't until I was pointed in the direction of [some Unix history](https://idea.popcount.org/2016-11-01-a-brief-history-of-select2/) that I learned what `select` is supposed to do.

I would like to propose that `select` is slightly renamed to make it more clear what it is actually doing, to e.g. `select_any` or `select_any_ready` maybe (or any other name that is deemed better). I know this breaks somewhat with the precedent for this functionality (since it is named as such in Unix and Go for example), but I think there are many Rust users who are not familiar with these precedents and will hence be confused by the name (myself included). At the same time, by keeping the word "select" in the name, it is still easy to search for, for those that do know the precedent (and if "select" is kept as the first part of the name, anyone just looking to type `select` in their IDE will also be helped by autocomplete).

I feel this would be a very rustic improvement to make; it should still be easy for veterans to use, but welcomes newcomers at the same time, by optimizing for readability over familiarity.

ghost:
@TimNN 

> I don't think there is an easy way for optional commas, but allowing an arbitrary number of commas ($(,)*) is easily possible (I've updated the gist).

Wonderful! Although, it's a bit surprising that the following code compiles even without braces and commas:

```rust
fn _foo(mut s: String, tx: Sender<String>, rx: &Receiver<String>) -> String {
    select! {
        disconnected() =>
            s
        // recovers `s` if unsuccessfull (currently needs `s` to be mutable,
        // but that can probably be worked around in the macro)
        send(tx, s) =>
            String::from("I did send s!")
        // Re-evaluates (and drops the result) on every execution
        send(tx, String::from("foo")) =>
            s
        recv(rx, val) =>
            val
    }
}
```

Can we make it an error if you omit both braces and commas?

> The `send(tx, $ident)` form of the current macro already recovers the value of `$ident` in case of an error (although that currently requires `$ident` to be `mut`, but that should be easy to fix: The value of `$ident` moves into the loop anyway and is unusable afterwards, so adding an `let mut $ident = $ident` before the loop should be easy enough)

Hmm. This wouldn't work if you're trying to send a field in a struct. For example:
```rust
select! {
    send(tx, foo.message) => { ... }
}
```

I'd still prefer to mark explicily that ownership must be reacquired, so that `send(tx, mut foo.message)`  expands to:
```rust
if let Err(m) = sel.send(tx, foo.message) {
    foo.message = m;
} else {
    // ...
}
```

One more thing. I'm afraid that the macro obscures the fact that it is actually a loop behind the scenes. A user writing the following might assume that `compute()` gets called just once.

```rust
select! {
    send(tx, compute()) => println!("sent!"),
    recv(rx, val) => println!("received {}", val),
}
```

If `compute()` has side-effects, the user might be in for a bad surprise. Perhaps we should rename the macro to `select_loop!` or `loop_select!`?

ghost:
@johncf 
> Here is a refinement of the idea I proposed before, now having APIs for sending and possibly recovering failed-to-send messages.

A few issues I can think of:

1. It's unfortunate that we have to mention each case twice - e.g. first as `.join(&rx2)` and then as `if let Some(msg) = select.recv(&rx2)`.
2. We have to write `else { unreachable!() }` at the end to make sure we didn't forget a case, or to make code typecheck in case we're using the final value of the whole if-elseif expression.

johncf:
@stjepang I have a few more concerns regarding a `select!` macro.

1.  You can't `break` or `continue` a loop wrapping the `select!` without using a label. (Also, `loop_select!` sounds like a helper macro for this use-case.)
2.  ~~To guarantee fairness, wouldn't it be necessary to use thread-local variables to store past state? Unless `select!` takes a local variable given by the user to store state, as follows:~~

    ```rust
    fn _foo(rx1: &Receiver<String>, rx2: &Receiver<String>) {
        let mut state = SelectState::new();
        loop {
            select! {state,
                disconnected() => {
                    break; // oops! ignore for now
                }
                recv(rx1, val) => {
                    if val == "exit" {
                        break;
                    } else {
                        System.out.println("Got {}", val);
                    }
                }
                recv(rx2, val) => {
                    System.out.println("Got {} from rx2", val);
                }
            }
        }
    }
    ```

3.  Is it possible to easily recover a variable passed into `send`? One possibility is:

    ```rust
    let (s_opt, recv_opt) = select! {
        disconnected() => {
            (Some(s), None)
        }
        send(tx, s) => {
            (None, None)
        }
        recv(rx, val) => {
            (Some(s), Some(val))
        }
    };
    System.out.println("{:?}", s_opt, recv_opt);
    ```

_Edit:_ Possible recovery of a sent variable on failure.\
_Edit2:_ Strikeout 2.

TimNN:
@johncf:

1) I think this is an acceptable downside. (You can statically prevent unlabeled `break`s / `continue`s inside the select macro).
2) The `SelectState::new()` stuff would be generated by the macro.
3) The recovery as written in your updated code works already with the macro I proposed (if `s` is `mut`, although that restriction can be lifted). [You original code example would produce a compiler error].

---

@stjepang:

> Can we make it an error if you omit both braces and commas?

Probably, but IMO that would complicate the macro a lot for very little gain.

johncf:
@TimNN 

1. Yes, I think I remember reading that. But will the compiler be able to provide a helpful message on how to fix that?
2. Whatever generated by the macro would be _inside_ the outer loop (see my code), but the state needs to be _outside_ to outer loop for persistence.
3. Yes, and imagine having one more `send`. I just think that it is a lot of inconvenience, for an "elegant" solution (which it is only for the very basic use-case).

TimNN:
@stjepang: I personally like `select_loop!`.

Regarding the `send()` variants, maybe three version would work:

1. `send($tx, $ident)`: The IMO most common case: `$ident` is made mutable and automatically recovered
2. `send($tx, mut $($ident).*)`: The (nested) struct field case (works for a single variable as well): `$ident` is assumed to be mutable and automatically recovered
3. `send($tx, eval $expr)`: An arbitrary expression, which is evaluated every time. (The `eval` key tries to highlight that `$expr` is evaluated multiple times)

TimNN:
@johncf 

2. Oh, I think I see what you mean. If I understand the RFC correctly you wouldn't need past state because the "start" case of the select loop is randomly determined each time the loop is initialized.

TimNN:
@johncf:

1. The error message would look something like this (which I believe is helpful enough): 

```
error[E0384]: re-assignment of immutable variable `_dont_use_an_unlabeled_continue_or_break_in_select_loop`
  --> src/main.rs:44:13
   |
44 |               _dont_use_an_unlabeled_continue_or_break_in_select_loop = ();
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ re-assignment of immutable variable```
```

3. Oh I think I misunderstood your point, did you mean if there was an easy way to recover the value to a send *after the select loop*? If that was the question, then I don't believe there is an easy way to do that without a significantly more complex / specialized / restricted macro.

johncf:
@TimNN 

1. 😕 hmm... Incorrect line numbers... I dunno YMMV.
2. Ah sorry, I think you're right. I only skimmed through that part earlier.
3. Yes, after the select-loop was what I meant. I'm not sure how common using `send` in `select` would be to comment about it though.

bstrie:
Reminder that potential edge-case shortcomings of the macro form aren't the end of the world, as long as it's documented how to use this library without the macro, via manually looping and breaking. If the macro will keep people from shooting themselves in the foot for 95% of use cases, then it's still a worthwhile addition.

ghost:
What if we simply add a new method `Select::is_done` that returns `true` when selection is done?
Then you don't have to manually break the loop in every successful case:

```rust
let mut sel = Select::new();
while !sel.is_done() {
    if let Ok(peer) = sel.recv(rx) {
        println!("{} received a message from {}.", name, peer);
    }
    if let Ok(()) = sel.send(tx, name) {
        // Wait for someone to receive my message.
    }
}
```

This way you can still put a `break` if you want, but it's not required.

TimNN:
@stjepang: I believe that would have the same downside as the proposed for-loop version: the compiler could no longer guarantee that each "branch" was only taken once. (And in that case I believe the for-syntax is a bit nicer).

text/2017-11-09-channel.md
ghost: I like the symmetry between `Result<(), T>` and `Result<T, ()>`.

If `recv` returns a `Result<T, SelectRecvError>`, then `send` should probably similarly return a `Result<(), SelectSendError<T>>`. That means we'd have to regain ownership from a failed `send` operation like this:

```rust
if let Err(err) = sel.send(&tx, name) {
    // Sending failed. Regain ownership of the message.
    name = err.into_inner();
} else {
    // Success! The message was sent.
    break;
}
```

Calling `.into_inner()` is slightly more unergonomic, but it's not too bad. Possible alternatives could be `.into()` and `.take()`. Do you agree with this reasoning? :)

ghost:
Thank you, everyone, for the feedback! I gave all this some thought over the last week.

Overall, it seems that the RFC is generally accepted quite well and there aren't any major blockers. I'll address some comments:

1. @droundy says selection is not exactly fair, which is true. I've added an item on this to the *Drawbacks* section. It's probably possible to make selection fairer, although it's a bit tricky to do correctly and efficiently. I'll open a separate issue on this.

2. @cramertj had a comment on the error type used in the `recv` selection case. I've introduced two new types to streamline the interface: `SelectSendError<T>` and `SelectRecvError`.

3. @matthieu-m had trouble understanding the selection mechanism in full detail. I've put some effort to clarify the text a bit (e.g. by explaining what wait lists are for). I know the text is not perfectly clear now either, but explaining all the subtle intricacies would require a lot more work. I'll probably put this effort into documenting the actual code instead. :)

4. The idea of switching from hidden thread-local state machine to explicit `Select` struct was universally accepted. I've addressed this in the text and in the implementation.

5. Some people wanted finer-grained control over `Sender` and `Receiver` by making them traits, or custom types for each channel variant (flavor). This is a reasonable idea, but I'd prefer to go with a simple interface that is basically a drop-in replacement for `std::sync::mpsc`, at least for now.

6. There were proposals for alternative selection interfaces. Unfortunately, I don't think any of those were tangibly better than the current one. All the proposals would fix some issues in the current mechanism, but then introduce new problems along the way.

7. A lot of people think selection has brittle interface (it's easy to accidentally use it incorrectly). I can agree with that. The best solution would be to introduce a macro for more convenient and foolproof selection. Thanks to @TimNN for a very promising prototype! I'm definitely onboard for a selection macro and will open an issue so we can discuss this further. A new RFC just for the macro may follow up later.

8. It's unfortunate that incorrect use of selection can lead to deadlocks and similar weirdness. We should definitely add some runtime checks to make sure that the same set of selection cases is visited in the same order in each iteration of the loop. Will open an issue for that.

Some bikeshedding... Channels are constructed using functions named `bounded` and `unbounded`, which should be pretty self-explanatory. I intentionally chose those names instead of `channel` and `sync_channel` (as in the standard library) to avoid confusion, especially because Go calls zero-sized channels *synchronous* and bounded channels *asynchronous* (while Rust calls those *synchronous* :D). How does everyone feel about that? A possible alternative would be to have just a single function `new(cap: Option<usize>)`.

Finally, I'd like to push forward and publish version 0.1.0 now so that everyone can start playing with the channel. The interface as presented in this RFC will not be set in stone, so if we figure out how to improve it down the line, we'll definitely do it! But let's get some mileage first. How does that sound, @jeehoonkang?

jeehoonkang:
- in order to serve as many people as possible, how about shipping @TimNN's macro in 0.1.0?  From the discussions so far, I couldn't see any drawback in doing so.

- I feel "bounded/unbounded" are better names than "sync/async"; they explain exactly what they do.  Also, I don't see much benefit of `new()` at this moment.  Let's hear experience from users after releasing it.

- Even though channel's code is not reviewed at all, I'm quite sure that it's well-written 😃, and at the very least it's tested and its interface is reviewed.  Also, it seems users are waiting for a new Crossbeam release a lot!  Let's review and enhance it after releasing it.

Thanks for this great job, let's move forward!

TimNN:
> * in order to serve as many people as possible, how about shipping @TimNN's macro in 0.1.0 From the discussions so far, I couldn't see any drawback in doing so.

I can send a PR with the initial macro to the prototype repo, which should also allow us to better discuss the implementation details there.

TimNN:
Regarding the fairness, I just had the following idea (I haven't properly thought it through yet, but wanted to write it down):

This approach is probably a bit more expensive but should still be manageable: The core idea is to change the "two half" iteration to something different, so that neighboring operations are not necessarily directly next to each other.

This can be achieved by, instead of choosing a split point, choosing a skip count: Chose a skip count X, then skip X operations, perform one operation, skip X operations, perform one operation and repeat until every operation has been performed once (but see caveats / details below).

 Example:

Consider six operations A - F. With a split point the "check" order would be something like C.D.E.F.A.B for a split point of 2.

For skip counts between 0 - 5 the following orders are possible: A.B.C.D.E.F, B.D.F.A.C.E, C.F.B.E.A.D, D.A.E.B.F.C, E.C.A.F.D.B and F.E.D.C.B.A.

Caveats / Details:
Assume `N` operations in the loop. To guarantee that we visit all operations, we need to find a prime `P > N` and assume there are actually `P` "operations" in the loop. We also choose a random skip count `X` (`0 <= X < N`). To visit all elements in the loop we then do the following: 

0) Start at position `i = 0`
1) Skip X operations (`i = (i + X) % P`)
2) If the current operation does not actually exist (`i >= N`) goto 1.
3) Visit the operation at position `i`
4) If we have not yet visited `N` operations, goto 1.

Since `P` is prime and `X < P`, we are guaranteed to visit every position once before repeating.

Fairness: If I remember my probability theory lessons correctly this system should guarantee that the probability that a specific operation B is visited directly after an operation A should be approximately equally likely for all possible operations B. This should provide some fairly strong fairness guarantees.

Disadvantages:
This method is less performant than the "split point" algorithm:

1) It requires finding a suitable prime `P`, although that should be trivial for any reasonable value of `N`.
2) For large values of `X` the loop may need to be executed almost `N` times instead of only twice. This may be especially problematic in the face of expensive send value computations / moves due to send value recovery.

And now I've spent way too long writing this down and it is way too late, and I just realized that this turns the whole operation essentially from O(n) to O(n^2), which is probably too slow for a default but maybe works as an "extra-fair" select. Anyway, I'll think about this some more tomorrow. Comments / Feedback are very welcome.

bstrie:
I'm in favor of `bounded`/`unbounded`.

droundy:
@stjepang Thanks for your thoughtful response and thorough summary.  I think your response to the unfairness issue was particularly fair.  (Apologies for the pun.)

On the bikeshedding, I agree that `bounded` and `unbounded` are way better than `sync`/`async`, but wonder whether even better names are possible.  In particular, I wonder whether 3 names could be more instructive, even if one name is a special case for `bounded(0)`.  To my mind a zero-length bounded channel is a distinctively different thing (in use) than a channel with a finite-length buffer, and I wonder whether there is a more descriptive name for it.  The zero-length buffer really means that the receiver and sender are synchronized.  In terms of reasoning (and deadlocks) this is very different behavior from a channel with either finite (but non-zero) or infinite buffer...

What about `blocking` (zero-length buffer), `buffered`, and `nonblocking`? That avoids the "sync" syllable which is backwards between go and standard rust, and focuses instead on whether the channel will block.  Of the three names, `buffered` doesn't quite match the others, which I don't like.  But a finite buffer will sometimes wait for input (or a reader) and will sometimes not block.

droundy:
@TimNN I think a simpler version of your algorithm would simply be to randomly choose one operation to attempt each time through the loop.  It is still O(N^2), but is easier to reason about than the skipping version.  The prefactor on the O(N^2) shouldn't be too bad if no extra work is done in the loop, an you *really* want no extra work to be done in the loop, or the implementation leaks.

Also O(N^2) isn't too bad if N is small, and when N is large is precisely when the non-fairness gets large with the current implementation, so users might be likely to value it more when it is more expensive.  That said, I think the current plan to document the unfairness and postpone any idea of fixing it is good.

TimNN:
> That said, I think the current plan to document the unfairness and postpone any idea of fixing it is good.

 I agree!

> I think a simpler version of your algorithm would simply be to randomly choose one operation to attempt each time through the loop.

Ah, now I understand your earlier comment. I don't think that could easily work, consider this because you aren't guaranteed to visit all operations once:

* No operation is ready, the thread parks itself.
* The thread is unparked
* The thread randomly visits N operations
* The operation that unparked the thread is *not* visited
* <I can imagine multiple things that happen now, but overall I would consider this situation bad>

TimNN:
I think adding the following API to `Select` would make using it (with the macro) even safer for negligible costs:

* `.enable_extra_checks()`: Called before the loop and enables extra checks (= the one below)
* `.register_loop_end()`: Called at the very end of the loop

This will allow checking if the same end of a specific channel was used more than once, which means that the macro can enforce / catch violations of all three "important rules one must follow in order to use selection correctly" listed in the RFC either at run- or compile-time.

**Edit:** Well, mostly, even with the macro one can do something like `select_loop! { send(pick_random_tx(), value) => ... }`, but that basically means someone is actively trying to break things.

TimNN:
Another thought: If this can easily be supported by the implementation it might be nice to have a `Select::send` version which takes a closure instead of a value: This version would attempt to reserve a spot in the channel and only if that is successful would it evaluate the closure.

droundy:
@TimNN You are right about my idea, it would only work if the thread never parked, which would be terribly inefficient.  Perhaps just creating an array with length of the number of options, and then shuffling it to determine the order of attempts? The shuffling is an O(N) operation, and then we'd have to run through the loop O(N) times, so it's still an O(N^2) approach, I think.

Regarding your closure suggestion, that would seem to make the code far easier to use without side effects or repeated computation.

ghost:
**PSA:** To everyone interested in the `select_loop!` macro: follow the discussion in [this pull request](https://github.com/stjepang/channel/pull/3).

ghost:
@droundy @TimNN  Regarding fairness, we could do the following.

In the first iteration of the loop, while counting cases we also check which cases are ready by calling internal methods `.can_send()` and `.can_recv()`. These methods check whether there's a waiting receiver/sender on the other side of the channel, respectively.

Typically we select over no more than a dozen or so cases, so we can remember which cases are ready by marking a bit for each ready case in a `u32` bitmask. Then we choose a random number between 0 and `bitmask.count_ones()` (let's call it R) and set `start` to the index of the Rth bit in the bitmask.

There is some performance cost associated with checking which cases are ready (I believe the cost is around 10%), but it's probably not too bad.

@droundy 
> What about `blocking` (zero-length buffer), `buffered`, and `nonblocking`?

For me, names `blocking`/`nonblocking` are not much more understandable than `synchronous`/`asynchronous`. :) I prefer to differentiate channels by how many messages they can hold: are they infinite, do they have a bounded buffer, or are they zero-sized. The last kind is very special, that is true. It is sometimes also called a *rendezvous* channel.

It might be nice to have a dedicated constructor for zero-sized channels, but I'm afraid at the same time it might also make constructors more complicated than is necessary.

Note that there is already some precedent for constructors similar to `bounded(cap)`. For example, in Go you can construct bounded channels with `make(chan int, 0)` and `make(chan int, 50)`, while in Rust you can do `sync_channel(0)` and `sync_channel(50)`.

ghost:
I'm merging this RFC and publishing the first version.
New repository: https://github.com/crossbeam-rs/crossbeam-channel
Documentation: https://docs.rs/crossbeam-channel

You can add `crossbeam-channel = "0.1.0"` to your `Cargo.toml` and try it out!

```rust
extern crate crossbeam_channel;

use crossbeam_channel::unbounded;

fn main() {
    let (tx, rx) = unbounded();
    tx.send("Hello").unwrap();
    tx.send("World").unwrap();
    drop(tx);

    for msg in rx {
        print!("{} ", msg);
    }
    println!();
}
```

I decided to include `select_loop!` macro for easier selection. Thanks again to @TimNN for implementing it! As is often the case with macros, there are still a few warts in it and you can run into weird compilation errors, but I believe it's much less likely than with `select!`.

Let me know if you have any feedback or comments. I'm especially curious:

* How difficult is it to switch from `std::sync::mpsc` to `crossbeam-channel`?
* Did you notice any difference in performance?
* How does the `select_loop!` macro feel?
* Did you notice weird compilation errors produced by `select_loop!`? Any examples?
* If you notice any bugs, please open an issue in the repo. Thanks!

Finally, the remaining problems can be tracked in dedicated issues:

* [Selection fairness](https://github.com/crossbeam-rs/crossbeam-channel/issues/1)
* [Runtime checks for selection](https://github.com/crossbeam-rs/crossbeam-channel/issues/2)

ghost:
Just one more bikesheddy question. :)

I'd like to split the `disconnected()` case into two: `all_disconnected()` (fired when all selected channels are disconnected) and `any_disconnected()` (fired when any selected channel is disconnected). Crate `ipc-channel` (used in Servo) needs this feature.

Do you think names `all_disconnected` and `any_disconnected` are adequate? Any better suggestions?

So the full list of possible selection cases would be: `send`, `recv`, `would_block`, `timed_out`, `any_disconnected`, `all_disconnected`.

(also, if you think `would_block` and `timed_out` should be renamed, please say so)

Issue #23, opened 2017-11-10
Custom collector in Deque


ghost:
In PR #21, @jeehoonkang and @Vtec234 pointed out that it should be possible to specify a custom `Collector` in the deque constructor (`Deque::new`).

This is not a high-priority issue at the moment, but let's kick off a discussion - how would we go about implementing such a feature?

ghost:
Here's one idea off the top of my head.

Custom collectors remind me very much of custom allocators ([`Alloc`](https://doc.rust-lang.org/nightly/alloc/allocator/trait.Alloc.html) trait), custom thread pools ([`ThreadPool`](https://docs.rs/rayon/0.9.0/rayon/struct.ThreadPool.html) in Rayon), custom [panic handlers](https://doc.rust-lang.org/std/panic/fn.set_hook.html), etc. There is usually an implicit, global default that covers 99% of use cases, but sometimes you still want to specify a custom one.

We could introduce a `GetHandle` trait (similar to `Alloc` trait) with a default implementor (`DefaultGetHandle` is similar to [`Heap`](https://doc.rust-lang.org/nightly/alloc/heap/struct.Heap.html)):

```rust
trait GetHandle {
    fn get_handle(&self) -> Handle;
}

struct DefaultGetHandle;

impl DefaultGetHandle {
    fn get_handle(&self) -> Handle {
        epoch::handle()
    }
}
```

Then we can implement `Deque` and `Stealer` like this:

```rust
struct Deque<T, GH: GetHandle = DefaultGetHandle> {
    // ...
    collector: GH,
}

impl<T, GH: GetHandle> Deque<T> {
    fn with_collector(get_handle: GetHandle) -> Deque<T> {
        Deque {
            // ...
            get_handle,
        }
    }
}

impl<T, GH: GetHandle> Stealer<T> {
    fn steal(&self) -> Steal<T> {
        let handle = self.deque.get_handle.get_handle();
        let guard = handle.pin();

        // ...
    }
}
```

Finally, let's create a custom collector and give to a newly constructed `Deque`:

```rust
lazy_static! {
    static ref MY_HANDLE: Collector = Collector::new();
}

thread_local! {
    static MY_HANDLE: Handle = MY_HANDLE.handle();
}

struct MyGetHandle;

impl GetHandle for MyGetHandle {
    #[inline]
    fn get_handle(&self) -> Handle {
        MY_HANDLE.with(|h| h.clone())
    }
}

fn main() {
[...]

PR #31, opened 2018-04-07
Global and local pinning

ghost:
Introduce the concept of *global pinning*, which allows one to pin the current thread without going through a thread-local handle.

In addition to that, we'll slightly tweak the general interface of `crossbeam-epoch` in order to fix several reported problems with it.

[Rendered](https://github.com/stjepang/rfcs-crossbeam/blob/global-and-local-pinning/text/2018-04-08-global-and-local-pinning.md)

I'm feeling torn between all the possible alternatives and don't expect the RFC to land in the original form. It's hard to choose the right solution here, so help me out. :)

Related issues:

* https://github.com/crossbeam-rs/crossbeam-epoch/issues/73
* https://github.com/crossbeam-rs/crossbeam-epoch/issues/71
* https://github.com/crossbeam-rs/crossbeam-skiplist/issues/3
* https://github.com/crossbeam-rs/rfcs/issues/23

@Amanieu This RFC is the answer (a belated one, sorry!) to your questions about *anonymous pinning*.

### DIFF
--- a/text/2018-04-08-global-and-local-pinning.md
+++ b/text/2018-04-08-global-and-local-pinning.md
@@ -112,7 +112,7 @@ Instead, we'll add functions `epoch::pin_fence()`, `Handle::pin_fence()`, and `C
 
 # Alternatives
 
-### Add safe `Guard::pin_with()` and remove `Guard::clone()`
+### Add safe `Handle::pin_with()` and remove `Guard::clone()`
 
 We can add a safe alternative to `Guard::pin()`:
 
--- a/text/2018-04-08-global-and-local-pinning.md
+++ b/text/2018-04-08-global-and-local-pinning.md
@@ -1,0 +1,247 @@
+# Summary
+
+Introduce the concept of *global pinning*, which allows one to pin the current thread without going through a thread-local handle.
+
+In addition to that, we'll slightly tweak the general interface of `crossbeam-epoch` in order to fix several reported problems with it.
+
+# Motivation
+
+The following problems with the current `Collector`/`Handle`/`Guard` interface have been reported:
+
+1. Users might mistakenly expect `Handle::clone()` to register a brand-new handle rather than returning a reference to the same handle. The method can be dangerously misleading.
+
+2. Handles are not `Send`. This is an annoying problem and it's not obvious how to cheat around it correctly when you really have to.
+
+3. Handles are not `Sync`. Sometimes it'd be nice to call `pin()` without registering a handle for each thread. In `SkipMap`, we need to pin inside `self.collector`, but creating a handle per-thread *and* per-instance is difficult. In `elfmalloc` there are situations where having to register handles is a nuisance.
+
+4. `epoch::pin()` will panic if the thread-local handle is destroyed. Fortunately, `LocalKey::try_with` will be soon stabilized in Rust 1.26. But how do we work around the panics then - what should we do exactly in case the handle is destroyed?
+
+# Detailed design
+
+### Global pinning
+
+Let's introduce something I've been referring to as *anonymous pinning* in the past, and has been briefly explained in [this comment](https://github.com/crossbeam-rs/crossbeam-epoch/pull/21#discussion_r144353198).
+
+The idea is to have two atomic counters in the `Collector` (more concretely, in the `Global` struct). They count how many `Guards` are currently pinned in the global epoch and the epoch preceding it:
+
+```rust
+pin_counters: [AtomicUsize; 2],
+```
+
+We introduce a new method `Collector::pin()`, which is just like `Handle::pin()`:
+
+```rust
+impl Collector {
+    fn pin(&self) -> Guard;
+}
+```
+
+We'll have to distinguish between guards pinned globally and locally. Since we're turning `Guard` into an `enum`, let's also add a special case for the unprotected guard:
+
+```rust
+enum Guard {
+    Global(*const Global),
+    Local(*const Local),
+    Unprotected,
+}
+```
+
+Now, the `Global::pin()` method might be implemented like this:
+
+```rust
+fn pin(&self) -> Guard {
+    // 0 if the epoch is even, or 1 if the epoch is odd.
+    let parity = self.epoch.load(Ordering::Relaxed).parity();
+
+    self.pin_counters[parity].fetch_add(1, Ordering::Relaxed);
+    atomic::fence(Ordering::SeqCst);
+
+    Guard::Global(self)
+}
+```
+
+Of course, unpinning will simply decrement the counter.
+
+**@jeehoonkang** [raised a concern](https://github.com/crossbeam-rs/crossbeam-epoch/pull/21#discussion_r144719080) with correctness of this method of pinning.
+
+Consider this. What happens if we read a stale value of `self.epoch` and increment a counter just after the epoch advanced one step forward, and execute the fence just before the epoch is advanced one step forward again? That means we're effectively two epochs behind, while other threads might think we're zero epochs behind.
+
+I think the problem is easily fixable by changing when a bag becomes expired. Rather than defining a bag is expired if it's at least 2 epochs old, we should say it's expired if it's at least 3 epochs old. That's it.
+
+Finally, an important insight ought to be made about the difference between global (through `Collector`) and local (through `Handle`) pinning. Think of `Collector::pin()` as the *simple & easy* way of using the GC, and consider `Handle::pin()` just an optimized variant that avoids contention by writing to a special thread-local structure (a handle). In that sense, handles are a completely optional optimization, nothing more!
+
+### Adjustments to handles
+
+Let's remove `Handle::clone()` because it's misleading. It hasn't proved to be very useful so far, and is only used by `epoch::default_handle()`. Moreover, `epoch::default_handle()` hasn't been useful either so let's remove it, too.
+
+On the other hand, `epoch::default_collector()` really is useful, and is used by `SkipMap` to initialize `self.collector` (of type `Collector`) to the default collector.
+
+The omission of `epoch::default_handle()` isn't a big problem since we already have `epoch::pin()` and `epoch::flush()`. Note that `epoch::pin()` is a much better option than `epoch::default_handle().pin()` anyway - it can automatically fall back to global pinning if the thread-local handle is destroyed:
+
+```rust
+fn pin() {
+    HANDLE.try_with(|h| h.pin()).unwrap_or_else(|| COLLECTOR.pin())
+}
+```
+
+Next, we implement `Send` for `Handle` and mark `Handle::pin()` as unsafe:
+
+```rust
+unsafe impl Send for Handle {}
+
+impl Handle {
+    unsafe fn pin(&self) -> Guard;
+}
+```
+
+The caller of `Handle::pin()` must pinky promise that the handle will not be sent to another thread while the `Guard` (or its clones) is alive. Unless, of course, the guard (and its clones) is simultaneously sent to the same thread, too.
+
+### Remove `is_pinned()`
+
+Another small tweak we'll do is remove `epoch::is_pinned()` and `Handle::is_pinned()`. These functions are only used by `crossbeam-deque` to make sure a fence is executed even if the thread is already pinned.
+
+Instead, we'll add functions `epoch::pin_fence()`, `Handle::pin_fence()`, and `Collector::pin_fence()`, which are equivalent to `pin()`, except they guarantee to execute a `SeqCst` fence.
+
+# Drawbacks
+
+* `epoch::default_handle()` is gone. But it doesn't seem to be useful and we have `epoch::pin()` and `epoch::flush()` anyway. We can reintroduce it in the future if a need actually comes up.
+
+* `Handle::clone()` is gone. However, nobody is using it and when it is used, it's used mistakenly. In short, it comes with practically zero benefits and a serious drawback.
+
+* `Handle::pin()` is now unsafe. See the *Alternatives* section for what we can do about it.
+
+# Alternatives
+
+### Add safe `Guard::pin_with()` and remove `Guard::clone()`
+
+We can add a safe alternative to `Guard::pin()`:
+
+```rust
+impl Handle {
+    fn pin_with<F, R>(&self, f: F) -> R
+    where
+        F: FnOnce(&mut Guard) -> R;
+}
+```
+
+This function is safe because it borrows the handle for the duration of pinning, which means it cannot be sent to another thread in the meantime. However, the user mustn't be able to clone the `Guard` and send the handle to another thread afterwards, as in this example:
+
+```rust
+let guard = handle.pin_with(|guard| guard.clone());
+thread::spawn(move || {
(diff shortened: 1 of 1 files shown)


Amanieu:
I'm pretty happy with this RFC. My use case is a bit complicated because it involves threads that can potentially be killed at any point. I've included a rough code example which shows how my code works:

```rust
// There is one of these for each thread. They are tracked in a global list of
// all threads. ThreadData must be Send since it may be dropped in a different
// thread when a thread is killed.
struct ThreadData {
    handle: Handle,
    guard: Option<Guard>,
}

// This what the main loop of a thread looks like. The key fact to remember here
// is that threads can be asynchronously killed at any point, except when the
// ThreadData is locked.
fn run_thread(thread_data: &Mutex<ThreadData>) {
    let guard;

    {
        // Lock the thread. This prevents our thread from getting killed.
        let lock = thread_data.lock();

        // Pin the thread by setting thread_data.guard to Some(Guard).
        let guard = lock.guard.get_or_insert_with(|| lock.handle.pin());

        // We cheat a bit here by re-binding the lifetime of the guard so that
        // it outlives the lock. This is safe because we know that one of the
        // following occurs:
        // - Our thread is not killed, and we don't touch thread_data.guard
        //   anywhere except at the end of this function.
        // - If our thread is killed then our killer will drop our ThreadData
        //   (which it must lock before killing us). This will unpin the thread
        //   and avoid any memory leaks.
        guard = unsafe { &*(guard as *mut Guard) };

        // The thread stays pinned even after we release the lock.
        drop(lock);
    }

    // The main body of the thread goes here. We can safely access lock-free
    // data structures using our Guard, while still allowing our thread to be
    // killed at any point.

    // Unpin the guard after we are done using it.
    thread_data.lock().guard = None;
}
```

Amanieu:
I don't care much for global pinning for my use case. There is one piece of code where I do need it but I've just used a `Mutex<Handle>` as a workaround (the mutex is needed for other reasons anyways, so performance is not an issue).

Amanieu:
I would add another alternative: only allow a single `Guard` to exist per `Handle` at any point. This allows `Handle` to be safely sent to another thread since you can't get another `Guard` out of it while the previous `Guard` is still alive.

This would be ideal for my use case since I only have a single `Guard` in use at any one point, however I understand that it may be too limiting as a general-purpose solution.

text/2018-04-08-global-and-local-pinning.md
```diff
@@ -0,0 +1,247 @@
...
+
+```rust
+enum Guard {
+    Global(*const Global),
+    Local(*const Local),
+    Unprotected,
+}
+```
+
+Now, the `Global::pin()` method might be implemented like this:
```
glaebhoerl, line 49: Did you mean `Collector::pin()`?

text/2018-04-08-global-and-local-pinning.md
```diff
@@ -0,0 +1,247 @@
...
+
+* `epoch::default_handle()` is gone. But it doesn't seem to be useful and we have `epoch::pin()` and `epoch::flush()` anyway. We can reintroduce it in the future if a need actually comes up.
+
+* `Handle::clone()` is gone. However, nobody is using it and when it is used, it's used mistakenly. In short, it comes with practically zero benefits and a serious drawback.
+
+* `Handle::pin()` is now unsafe. See the *Alternatives* section for what we can do about it.
+
+# Alternatives
+
+### Add safe `Guard::pin_with()` and remove `Guard::clone()`
```
glaebhoerl: Did you mean `Handle::pin_with()`?

text/2018-04-08-global-and-local-pinning.md
ghost: Well, `Collector::pin()` would simply call `Global::pin()`, just like `Handle::pin()` now calls `Local::pin()`. It's the same thing :)

text/2018-04-08-global-and-local-pinning.md
ghost: Yes, thank you! Fixed.

text/2018-04-08-global-and-local-pinning.md
glaebhoerl: (Oh, fsr I hadn't realized `Global` is also a type and not only a variant.)

ghost:
> I would add another alternative: only allow a single Guard to exist per Handle at any point. This allows Handle to be safely sent to another thread since you can't get another Guard out of it while the previous Guard is still alive.

If we add a lifetime to `Guard`, then we could do something like this:

```rust
struct Handle { ... }

struct Guard<'a> { ...  }

impl Handle {
    fn pin(&self) -> Guard<'_> { ... }

    fn pin_mut(&mut self) -> Guard<'_> { ... }
}
```

This way you can obtain a unique guard if you want (`pin_mut`), but don't have to (`pin`). Is this what you had in mind?

Amanieu:
@stjepang That's doesn't work for me since I need to keep both the `Guard` and `Handle` in thread-local storage, which means that both must be `'static`.

ghost:
@Amanieu I see.

So in that case would you prefer something similar to [`raw_lock/raw_unlock`](https://docs.rs/parking_lot/0.5.4/parking_lot/struct.Mutex.html#method.raw_unlock) from `parking_lot`, except in our case the methods would be named `Handle::raw_pin` and `Handle::raw_unpin` (the latter one is `unsafe`).

What kind of interface would suit you best?

Amanieu:
Here is the interface that I have in mind:
```rust
impl Handle {
    // This fails if a Guard already exists for this Handle
    fn pin(&self) -> Option<Guard>;
}
```
`Guard` will no longer be `Clone`, which together with the above change, ensures that there is at most one `Guard` per `Handle` at any time. This will allow us to make both `Handle: Send` and `Guard: Send`.

Amanieu:
Only allowing one guard at a time might be too restrictive for the global collector since it doesn't allow nesting, so we can also provide this API only for the global collector (using the hidden per-thread handle):

```rust
thread_local! {
    static HANDLE: (Handle, Option<Guard>) = (COLLECTOR.register(), None);
}

pub fn epoch::pin_with<F: FnOnce(&Guard)>(f: F);
```

This will return a reference to the currently active thread-local `Guard` if there is one, or pin the epoch if there is no currently active guard.

ghost:
@Amanieu That seems a bit too restrictive - in particular, it would make implementing iterators that own guards impossible, wouldn't it?

However, going the other way around and implementing the *one-guard-per-handle* interface by wrapping the current one would be straightforward (except for implementing `MyGuard: Send`):

```rust
struct MyHandle {
    handle: Handle,
    pinned: Cell<bool>,
}

impl MyHandle {
    fn pin(&self) -> Option<MyGuard> {
        if self.pinned.get() {
            None
        } else {
            Some(self.handle.pin())
        }
    }
}

struct MyGuard {
    guard: Guard,
}

unsafe impl Send for MyHandle {}
```

Getting `MyGuard: Send` right is a bit tricky, though...

Maybe we should expose a few low-level methods for pinning without creating guards (like `raw_lock` and `raw_unlock`):

```rust
impl Handle {
    unsafe fn pin_unchecked();
    unsafe fn unpin_unchecked();
    unsafe fn defer_unchecked<F: FnOnce()>(f: F);
}
```

Then we can implement your interface like this:

```rust
struct MyHandle {
    handle: Handle,
    pinned: AtomicBool,
}

impl MyHandle {
    fn pin(&self) -> Option<MyGuard> {
        if self.pinned.swap(true, Relaxed) {
            None
        } else {
            unsafe { self.pin_unchecked(); }
            Some(MyGuard { handle: self.clone() }) // atomically increments the refcount
        }
    }
}

struct MyGuard {
    handle: Handle,
}

impl Drop for MyGuard {
    fn drop(&mut self) {
        unsafe { self.handle.pin_unchecked(); }
    }
}

unsafe impl Send for MyHandle {}
unsafe impl Send for MyGuard {}
```

Amanieu:
The problem with `MyHandle` is that it needs to give a `&Guard` because that is what `SkipList` expects. However this is unsound because the `Guard` can be cloned. This breaks the invariant of only having one guard per handle, which `MyHandle` requires to be able to be safely sent to another thread (the `Guard` must always be sent along with the `Handle`).

After consideration, I am happy with your proposed API except that I would like `Guard::clone` to be removed.

Amanieu:
For reference, here's my implementation of `SendHandle`: 
```rust
struct SendHandle {
    handle: Handle,
    guard: Option<Guard>,
}
unsafe impl Send for SendHandle {}
impl SendHandle {
    pub fn new(collector: &Collector) -> SendHandle {
        SendHandle {
            handle: collector.register(),
            guard: None,
        }
    }
    pub fn pin(&mut self) -> GuardRef {
        let handle = &mut self.handle;
        self.guard.get_or_insert_with(|| handle.pin());
        GuardRef(&mut self.guard)
    }
    pub fn get(&mut self) -> Option<&Guard> {
        self.guard.as_ref()
    }
    pub fn unpin(&mut self) {
        self.guard = None;
    }
}
struct GuardRef<'a>(&'a mut Option<Guard>);
impl<'a> GuardRef<'a> {
    pub fn forget(self) {
        mem::forget(self);
    }
}
impl<'a> core::ops::Deref for GuardRef<'a> {
    type Target = Guard;
    fn deref(&self) -> &Guard {
        self.0.as_ref().unwrap()
    }
}
impl<'a> Drop for GuardRef<'a> {
    fn drop(&mut self) {
        *self.0 = None;
    }
}
```

Two points of note:
- It only exposes a `&Guard`, not a `&mut Guard`. The latter is unsound because you could "steal" the guard with `mem::replace`.
- This implementation is unsound if `Guard` is clonable because you could send the `SendHandle` to thread B while still having a guard in thread A, which can lead to data races.

text/2018-04-08-global-and-local-pinning.md
```diff
@@ -0,0 +1,247 @@
...
+}
+```
+
+Of course, unpinning will simply decrement the counter.
+
+**@jeehoonkang** [raised a concern](https://github.com/crossbeam-rs/crossbeam-epoch/pull/21#discussion_r144719080) with correctness of this method of pinning.
+
+Consider this. What happens if we read a stale value of `self.epoch` and increment a counter just after the epoch advanced one step forward, and execute the fence just before the epoch is advanced one step forward again? That means we're effectively two epochs behind, while other threads might think we're zero epochs behind.
+
+I think the problem is easily fixable by changing when a bag becomes expired. Rather than defining a bag is expired if it's at least 2 epochs old, we should say it's expired if it's at least 3 epochs old. That's it.
```
Vtec234, line 69: Please forgive my ignorance, but as I understand, the benefit of using parity over a single counter is that with parity the epoch can still be advanced by 1 when there is a global pin present, right? If a larger counter is used, say `mod 4`, and then the epoch age required for destruction extended to 5, can the epoch advance by up to 3 with a global pin present? I'm not really proposing anything like that, but I'm curious about the relationship between the number of counters and the minimal free-able epoch age.

text/2018-04-08-global-and-local-pinning.md
```diff
@@ -0,0 +1,247 @@
...
+drop(guard); // Oops, the handle now belongs to another thread!
+```
+
+Therefore, we'd have to remove the `Clone` impl for `Guard`.
+
+### Make `Handle::pin()` safe, but add a lifetime to `Guard`
+
+Another alternative is to add a lifetime to `Guard` so that it becomes `Guard<'a>`. The lifetime `'a` borrows the `Handle` and therefore prevents it from being moved at all, which means it cannot be sent to another thread while such a guard exists.
+
+This way we don't have to remove `Handle::clone()`.
```
Vtec234, line 143: I like this alternative, but if we do go with it, I suggest to remove `Handle::clone()` anyway, since as you mentioned it is not clear what the method should do.

text/2018-04-08-global-and-local-pinning.md
```diff
@@ -0,0 +1,247 @@
...
+}
+
+// This function uses the default thread-local handle, but it cannot be sent to another thread,
+// so we don't even have to borrow it.
+fn pin() -> Handle<'static>;
+```
+
+The main drawback of this solution is that the lifetime might be at times annoying, but note that the lifetime can be simply elided in most real-world situations.
+
+### Two kinds of handles: `SharedHandle` and `LocalHandle`
```
Vtec234, line 165: Are there any drawbacks to this alternative besides it being somewhat more complex to implement and use?

Vtec234 (commented):
Left some questions, but in general looks very promising!

text/2018-04-08-global-and-local-pinning.md
stjepang: > Please forgive my ignorance, but as I understand, the benefit of using parity over a single counter is that with parity the epoch can still be advanced by 1 when there is a global pin present, right?

text/2018-04-08-global-and-local-pinning.md
ghost: Not really, that's the only drawback.

ghost:
@Amanieu

> After consideration, I am happy with your proposed API except that I would like Guard::clone to be removed.

That seems reasonable. One thing to keep in mind is that this precludes us from ever adding method `Guard::handle()` returning an `Option<&Handle>` (because one could just do `guard.handle().pin()` to clone it). But I suppose it's worth making that tradeoff.

@Vtec234

> Please forgive my ignorance, but as I understand, the benefit of using parity over a single counter is that with parity the epoch can still be advanced by 1 when there is a global pin present, right? 

Yes, exactly.

> If a larger counter is used, say mod 4, and then the epoch age required for destruction extended to 5, can the epoch advance by up to 3 with a global pin present?

Actually, it doesn't matter whether we have 2, or 4, or even 100 counters. Here goes a lengthy explanation... Sorry in advance - it'll require some mental gymnastics. :)

Consider what happens in our current epoch tracking mechanism. In order to pin the current thread, we (1) load the global epoch, (2) store it into our thread-local storage, and (3) execute a fence. Since `1 << 63` is so large, we can pretty safely assume that that the global epoch won't overflow between steps (2) and (3). That is just extremely unlikely to happen. :) One more thing: recall that our basic invariant is that the global epoch can advance at most **once** while the thread is pinned (i.e. between step (3) and the final call to `unpin`).

But what if it was likely for the global epoch to advance by `1 << 63` (on 64-bit platforms) between steps (2) and (3)? Let's say that in step (1) we load global_epoch=7. Now assume that after we store it into thread-local storage in step (2) and execute the fence in step (3), the global epoch almost wraps around and becomes global_epoch=6. Concurrently, another thread is executing function `try_advance` - it managed to iterate the whole list between our steps (2) and (3), and is now almost finished - the only thing left to do is to increment the global epoch. After our step (3) has finished, that thread sets global_epoch=8. This means that the global epoch can in fact advance **twice** while we're pinned (first from 6 to 7, and then from 7 to 8).

If we assume it's possible for the global epoch to advance by `1 << 63` between steps (2) and (3), then we have to change the expiration threshold from 2 to 3. The reason why we need to do that if we introduce global pinning with two counters is simply because it's very possible for global epoch to wrap around modulo 2 (in contrast to wrapping modulo `1 << 63`).

